PR #3, opened 2019-04-08
Add a new domain and implementation for ProgramWorkflow

ridmal:
- Add programWorkflow CSV parsers
- Add tests for programWorkflow loading
- Add CSV file to keep the programWorkflow meta-data


api/src/main/java/org/openmrs/module/initializer/InitializerConstants.java
```diff
@@ -51,4 +51,6 @@
 	
 	public static final String DOMAIN_PROG = "programs";
 	
+	public static final String DOMAIN_PROG_WF = "programWorkflows";
```
mks-d: The convention for the values here is 'no case' (look at the others), make it `"programworkflows"`:
```java
public static final String DOMAIN_PROG_FL = "programsworkflows";
```

api/src/main/java/org/openmrs/module/initializer/api/programWorkflow/ProgramWorkflowCsvParser.java
```diff
@@ -0,0 +1,46 @@
...
+	protected boolean isVoidedOrRetired(ProgramWorkflow instance) {
+		return instance.isRetired();
+	}
+	
+	@Override
+	protected ProgramWorkflow doVoidRetireActions(ProgramWorkflow instance) {
+		ProgramWorkflow programWorkflow = service.getWorkflowByUuid(instance.getUuid());
+		if (programWorkflow != null) {
+			Program program = instance.getProgram();
+			program.retireWorkflow(instance);
```
mks-d: At this point `instance` is already marked as retired and will be saved as such later down the line anyway. Are you not supposed to simply do:
```java
program.removeWorkflow(instance);
```
?

api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
```diff
@@ -160,6 +160,13 @@ protected void close() throws IOException {
 		reader.close();
 	}
 	
+	/*
+	 * Action for void/retire instances.
+	 */
```
mks-d: Make it a JavaDoc, and more explicit:
```java
/**
 * This method provides a space to perform actions after setting an object as voided or retired.
 *
 * Eg.: A program workflow needs to be removed from its program after being marked as retired. 
 * @see ProgramWorkflowCsvParser#doVoidRetireActions(org.openmrs.ProgramWorkflow)
 */
```

api/src/main/java/org/openmrs/module/initializer/api/programWorkflow/ProgramWorkflowLineProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+import org.openmrs.api.context.Context;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.openmrs.module.initializer.api.impl.Utils;
+
+public class ProgramWorkflowLineProcessor extends BaseLineProcessor<ProgramWorkflow, ProgramWorkflowService> {
+	
+	protected static String HEADER_PROGRAM = "program";
+	
+	protected static String HEADER_CONCEPT = "concept";
```
mks-d:
```java
protected static String HEADER_WORKLOW_CONCEPT = "workflow concept";
```

api/src/test/java/org/openmrs/module/initializer/api/DomainProgramWorkFlowsInitializerServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.openmrs.test.BaseModuleContextSensitiveTest;
+import org.openmrs.test.Verifies;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+import java.util.HashSet;
+import java.util.Locale;
+import java.util.Set;
+
+public class DomainProgramWorkFlowsInitializerServiceTest extends DomainBaseModuleContextSensitiveTest {
```
mks-d: Just thinking, wouldn't you have a lot already available to help you run tests if you were to inherit `DomainProgramsInitializerServiceTest` and rerun its setup in this test's setup?
I'm not sure, you need to try. At least you could expose some test utils that one can reuse from the other.

api/src/test/java/org/openmrs/module/initializer/api/DomainProgramWorkFlowsInitializerServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+			Assert.assertEquals(cs.getConceptByName("concept1"), programWorkflow.getConcept());
+			Assert.assertEquals(false, programWorkflow.isRetired());
+			
+			Program program = pws.getProgramByName("program1");
+			Assert.assertEquals(programWorkflow, program.getWorkflow(programWorkflow.getId()));
+			
+		}
+		{
+			ProgramWorkflow programWorkflow = pws.getWorkflowByUuid("2b98bc76-245c-11e1-9cf0-00248140a5ef");
+			Assert.assertNull(programWorkflow);
```
mks-d: What is this test showing exactly?

api/src/test/java/org/openmrs/module/initializer/api/DomainProgramWorkFlowsInitializerServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+		// A Program used for add ProgramWorkflows
+		{
+			Program pro = new Program();
+			pro.setConcept(cs.getConceptByName("programConcept"));
+			pro.setOutcomesConcept(cs.getConceptByName("outcomesConceptTest1"));
+			pro.setName("program1");
+			pro.setDescription("Program Description");
+			pro = pws.saveProgram(pro);
+		}
+		// A programWorkFlow to be edited
```
mks-d: Do you mean?
>`// A workflow to be removed from its program`

api/src/test/java/org/openmrs/module/initializer/api/DomainProgramWorkFlowsInitializerServiceTest.java
ridmal: This is for showing voided programWorkflows are not adding to the database.

api/src/main/java/org/openmrs/module/initializer/api/programWorkflow/ProgramWorkflowCsvParser.java
ridmal: Actually first i used that. But in openMRS core , when we remove a programWorkflow from a program, it just changed as retired. But like you said it already done in the ```programWorkflow.setRetired(getVoidOrRetire(line.asLine()));``` we can skip that.

api/src/test/java/org/openmrs/module/initializer/api/DomainProgramWorkFlowsInitializerServiceTest.java
mks-d: Ok, we never were adding such tests until now, but then leave it ;-)

However we also need a real case of _edition_ to be tested, showing that changing the defining concept is well taken into account.

api/src/main/java/org/openmrs/module/initializer/api/programWorkflow/ProgramWorkflowCsvParser.java
mks-d: Can you detail exactly what each of
- `program.retireWorkflow(instance)`
- `program.removeWorkflow(instance)`

does?

PR #5, opened 2019-04-16
Add programWorkflowState domain

ridmal:


api/src/main/java/org/openmrs/module/initializer/api/Utils.java
```diff
@@ -211,7 +212,7 @@ public static LocationTag fetchLocationTag(String id, LocationService service) {
 	 * 
 	 * @param id The program UUID, name or underlying concept identifier (name, UUID or 'same as'
 	 *            concept mapping).
-	 * @return The {@link Program} instance if found, null otherwiwse.
+	 * @return The {@link Program} instance if found, null otherwise.
```
mks-d: Thanks!

api/src/test/java/org/openmrs/module/initializer/api/UtilsTest.java
```diff
@@ -75,4 +83,27 @@ public void getBestMatchName_shouldReturnBestMatchForConceptName() throws Except
 		Assert.assertEquals("A preferred name in English", Utils.getBestMatchName(c, Locale.FRENCH));
 		Assert.assertEquals("A preferred name in English", Utils.getBestMatchName(c, Locale.GERMAN));
 	}
+	
+	@Test
+	public void fetchProgram_shouldReturnProgramFromGivenId() throws Exception {
+		
+		ConceptService cs = Context.getConceptService();
+		ProgramWorkflowService pws = Context.getProgramWorkflowService();
+		ProgramsLoaderIntegrationTest.setupPrograms(cs, pws);
```
mks-d: You shouldn't really be using this here, `ProgramsLoaderIntegrationTest.setupPrograms(..)` is a routine intended for _integration_ tests (so tests using Spring Test), but here you are working on a _unit_ test.

Also we need to make a separate test case that asserts that null is returned when several programs are defined by the same concept.

PR #7, opened 2019-04-23
Privilege domain Implementation 

ridmal:
* Implementation of Privilege Domain.
* Add Integration test for privilege domain implementation.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
@@ -14,6 +14,7 @@ import org.apache.commons.logging.LogFactory;
 import org.openmrs.BaseOpenmrsData;
 import org.openmrs.BaseOpenmrsMetadata;
 import org.openmrs.BaseOpenmrsObject;
+import org.openmrs.Privilege;
 import org.openmrs.api.APIException;
 import org.openmrs.api.OpenmrsService;
 import org.openmrs.module.initializer.InitializerConstants;
@@ -175,7 +176,9 @@ public abstract class CsvParser<T extends BaseOpenmrsObject, S extends OpenmrsSe
 				T instance = createInstance(line);
 				if (instance != null) {
 					instance = save(instance);
-					if (instance.getId() != null) {
+					if (instance instanceof Privilege && ((Privilege) instance).getPrivilege() != null) {
+						instances.add(instance);
+					} else if (instance.getId() != null) {
 						instances.add(instance);
 					}
 				}
--- a/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
@@ -176,9 +176,7 @@ public abstract class CsvParser<T extends BaseOpenmrsObject, S extends OpenmrsSe
 				T instance = createInstance(line);
 				if (instance != null) {
 					instance = save(instance);
-					if (instance instanceof Privilege && ((Privilege) instance).getPrivilege() != null) {
-						instances.add(instance);
-					} else if (instance.getId() != null) {
+					if (isSaved(instance)) {
 						instances.add(instance);
 					}
 				}
@@ -191,5 +189,12 @@ public abstract class CsvParser<T extends BaseOpenmrsObject, S extends OpenmrsSe
 		} while (line != null);
 		
 		return instances;
+	}
+	
+	/**
+	 * Return true if instance is actually saved in database.
+	 */
+	protected boolean isSaved(T instance) {
+		return instance.getId() != null;
 	}
 }
--- a/api/src/test/java/org/openmrs/module/initializer/api/PrivilegesLoaderIntegrationTest.java
+++ b/api/src/test/java/org/openmrs/module/initializer/api/PrivilegesLoaderIntegrationTest.java
@@ -1,0 +1,87 @@
+package org.openmrs.module.initializer.api;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.Privilege;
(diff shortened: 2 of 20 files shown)


api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
```diff
@@ -175,7 +176,9 @@ protected void close() throws IOException {
 				T instance = createInstance(line);
 				if (instance != null) {
 					instance = save(instance);
-					if (instance.getId() != null) {
+					if (instance instanceof Privilege && ((Privilege) instance).getPrivilege() != null) {
+						instances.add(instance);
+					} else if (instance.getId() != null) {
```
mks-d: @ridmal we will certainly not single out privileges here at the level of the template class. Can you expand on what you encountered that you needed to do this?

ridmal: Actually Privilege instance don't have any id , because of that it keeps giving waring like this. 
https://hastebin.com/ubasotufuv.rb  
But test didn't fail without above implementation. :)

api/src/test/resources/testAppDataDir/configuration/privileges/privileges.csv
```diff
@@ -0,0 +1,6 @@
+Uuid,Name,Description
```
mks-d: Let's call this `Privilege Name` as it would show up on the legacy UI.

api/src/test/resources/testAppDataDir/configuration/privileges/privileges.csv
```diff
@@ -0,0 +1,6 @@
+Uuid,Name,Description
+494e1213-360d-45ac-877d-515993444290,Add Cohort,Able to add a cohort to the system
+cf687ee2-2700-102b-80cb-0017a47871b2,Add Concept Proposal,
+cf68b952-2700-102b-80cb-0017a47871b2,Edit Concepts,Able to change attributes of existing terms
+cf690132-2700-102b-80cb-0017a47871b2,,Allows users to manage programs
+cf688144-2700-102b-80cb-0017a47871b2,,Able to add new terms to the concept dictionary
```
mks-d: The principle of Iniz that each CSV line fully describes what will be saved in the end. The last two lines suggest that
- either a privilege is being created without a privilege name.
- or a privilege is being edited with its privilege name being set to an empty string.

Those cases should generate an error that is being logged.

mks-d: Privileges are a rare case where UUIDs won't matter that much, we will in practise rather always bring in lines like this:

| Uuid | Privilege name | Description                              | 
|------|----------------|------------------------------------------| 
|      | Add Attachment | Ability to add attachments to a patient. | 

You should definitely show that it is possible to create and edit a privilege that way.

ridmal: Yes last line will not added any data to database. But line before last one i included to show that we can edit description of a privilege even without privilege name. ( If user not able to put privilege name , in the current implementation it will find privilege from the UUID and update the description. )

mks-d: That line will edit the description based on the UUID that is provided yes, but **should also set the name to an empty string as per the CSV line**, and this is where an exception should be thrown.

api/src/test/java/org/openmrs/module/initializer/api/PrivilegesLoaderIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+		{
+			Privilege priv = us.getPrivilegeByUuid("cf688144-2700-102b-80cb-0017a47871b2");
+			Assert.assertNull(priv);
+		}
+		// unedited privilege
+		{
+			Privilege priv = us.getPrivilegeByUuid("cf687ee2-2700-102b-80cb-0017a47871b2");
+			Assert.assertNotNull(priv);
+			Assert.assertEquals("Add Test Privilege", priv.getName());
+			Assert.assertEquals("Add Test Privilege", priv.getPrivilege());
```
mks-d: After processing the CSV file this privilege should have
- name = "Add Concept Proposal"
- desc. = ""

How come it was left unedited?

api/src/test/java/org/openmrs/module/initializer/api/PrivilegesLoaderIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+			Assert.assertEquals("Edit Concepts", priv.getPrivilege());
+			Assert.assertEquals("Able to change attributes of existing terms", priv.getDescription());
+		}
+		{
+			Privilege priv = us.getPrivilege("Add People");
+			Assert.assertNotNull(priv);
+			Assert.assertEquals("Add People", priv.getName());
+			Assert.assertEquals("Able to add person objects", priv.getDescription());
+		}
+		// uncreated privilege
```
mks-d: `// CSV line without name should result in no privilege creation`

mks-d: I like that test! :-)

api/src/test/java/org/openmrs/module/initializer/api/PrivilegesLoaderIntegrationTest.java
ridmal: @mks-d  We cannot edit privilege name. ( So in the implementation it throws an exception when user try to do that)

mks-d: @ridmal ok then in the comment explain your assert:
`// a privilege name cannot be edited`

PR #10, opened 2019-05-02
OPENMRS-77: Added support for metadata mappings.

Ruhanga:
Ticket : https://mekomsolutions.atlassian.net/browse/OPENMRS-77


api/src/test/resources/org/openmrs/module/initializer/include/metadata/metadataMappingTermsDataSet.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
mks-d: Let's just create those in the setup of your integration test through using Metadata Mapping API.
I think test datasets are good if you intend to reuse them a lot across many tests (like it is done with the standard test dataset in Core). But here we do in fact isolate each domain and test it separately.

Don't get me wrong, this is not incorrect, but I would rather keep  making tests the same way then. Here you're opening the door to creating a full reusable dataset for the whole of Iniz, why not? But not now then.

api/src/test/resources/testAppDataDir/configuration/metadatamapping/metadataterms.csv
```diff
@@ -0,0 +1,4 @@
+Uuid,Void/Retire,MappingCode,SourceName,ClassName,MetadataUuid,-order:2000
```
mks-d: The headers should change, but also this is not correct: `-order:2000`. Just put it at `_order:1000`. See [here](https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/csv_conventions.md#header-_order) for details.

api/src/main/java/org/openmrs/module/initializer/InitializerConstants.java
```diff
@@ -57,4 +57,6 @@
 	
 	public static final String DOMAIN_PRIV = "privileges";
 	
+	public static final String DOMAIN_METADATA_MAP = "metadatamapping";
```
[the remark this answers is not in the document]
Ruhanga: Resolved

api/src/main/java/org/openmrs/module/initializer/api/Utils.java
```diff
@@ -270,6 +273,38 @@ public static ProgramWorkflowState fetchProgramWorkflowState(String id, ProgramW
...
 	}
 	
+	/**
+	 * Fetches a metadataterm for the id.
+	 * 
+	 * @param id The metadataterm UUID.
+	 * @param service
+	 * @return The {@MetadataTermMapping } instance if found, null otherwise.
+	 */
+	public static MetadataTermMapping fetchMetadataTermMapping(String id, MetadataMappingServiceWrapper service) {
```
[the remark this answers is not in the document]
Ruhanga: Logic migrated to `MetadataMappingServiceWrapper`.

api/src/main/java/org/openmrs/module/initializer/api/Utils.java
```diff
@@ -270,6 +273,38 @@ public static ProgramWorkflowState fetchProgramWorkflowState(String id, ProgramW
...
+	}
+	
+	/**
+	 * Fetches a metadataterm for the id.
+	 * 
+	 * @param sourceName The metadata source name.
+	 * @param service
+	 * @return The {@MetadataSource } instance if found, throws exception otherwise.
+	 */
+	public static MetadataSource fetcgMetadataSourceByName(String sourceName, MetadataMappingServiceWrapper service)
```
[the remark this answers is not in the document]
Ruhanga: Resolved.

api/src/main/java/org/openmrs/module/initializer/api/loaders/MetadataMappingsLoader.java
```diff
@@ -0,0 +1,35 @@
...
+	private MetadataMappingService service;
+	
+	@Override
+	public String getDomain() {
+		return InitializerConstants.DOMAIN_METADATA_MAP;
+	}
+	
+	@Override
+	public Integer getOrder() {
+		return 14;
```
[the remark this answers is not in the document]
Ruhanga: I have refactored the order to occur before GPs' loader.

api/src/main/java/org/openmrs/module/initializer/api/metadata/MetadataTermMappingLineProcessor.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class MetadataTermMappingLineProcessor extends BaseLineProcessor<MetadataTermMapping, MetadataMappingServiceWrapper> {
+	
+	protected static String SOURCE_NAME = "SourceName";
+	
+	protected static String MAPPING_CODE = "MappingCode";
+	
+	protected static String CLASS_NAME = "ClassName";
+	
+	protected static String METADATA_UUID = "MetadataUuid";
```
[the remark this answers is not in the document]
Ruhanga: Resolved.

api/src/main/java/org/openmrs/module/initializer/api/metadata/MetadataTermMappingLineProcessor.java
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	public MetadataTermMappingLineProcessor(String[] headerLine, MetadataMappingServiceWrapper service) {
+		super(headerLine, service);
+	}
+	
+	@Override
+	protected MetadataTermMapping bootstrap(CsvLine line) throws IllegalArgumentException {
+		String uuid = getUuid(line.asLine());
+		
+		MetadataTermMapping mapping = Utils.fetchMetadataTermMapping(uuid, service);
```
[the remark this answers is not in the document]
Ruhanga: Makes sense to me. Question is whether non of the `source, code` would need to be updated if exists.

api/src/test/resources/org/openmrs/module/initializer/include/metadata/metadataMappingTermsDataSet.xml
Ruhanga: Yes, I get the Idea here.

api/src/test/resources/testAppDataDir/configuration/metadatamapping/metadataterms.csv
Ruhanga: Yes, very true. I initially got test failures because of this.

api/src/main/java/org/openmrs/module/initializer/api/metadata/MetadataTermMappingLineProcessor.java
```diff
@@ -0,0 +1,55 @@
...
+		return mapping;
+	}
+	
+	@Override
+	protected MetadataTermMapping fill(MetadataTermMapping mapping, CsvLine line) throws IllegalArgumentException {
+		MetadataSource source = Utils.fetcgMetadataSourceByName(line.get(SOURCE_NAME), service);
+		mapping.setMetadataSource(source);
+		mapping.setCode(line.get(MAPPING_CODE));
+		mapping.setMetadataUuid(line.get(METADATA_UUID));
+		mapping.setMetadataClass(line.get(CLASS_NAME));
```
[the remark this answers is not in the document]
Ruhanga: Addressed in this commit https://github.com/mekomsolutions/openmrs-module-initializer/pull/10/commits/9eec778c15337ae11d705da4b306974c6a09cf8f

api/.gitignore
```diff
@@ -0,0 +1 @@
+/bin/
```
mks-d: Rather put this in the one top level **.gitignore**. Something like
```
/api/bin
```

api/src/main/java/org/openmrs/module/initializer/api/metadata/MetadataMappingServiceWrapper.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+	
+	/**
+	 * Fetches a metadataterm for the id.
+	 * 
+	 * @param sourceName The metada source name.
+	 * @param mappingCode
+	 * @return The {@MetadataTermMapping } instance if found, new instance otherwise.
+	 */
+	public MetadataTermMapping fetchMetadataTermMapping(String sourceName, String mappingCode)
```
mks-d: Make it just a getter:
```java
public MetadataTermMapping getMetadataTermMapping(String sourceName, String mappingCode)
```

api/src/main/java/org/openmrs/module/initializer/api/metadata/MetadataMappingServiceWrapper.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+	
+	/**
+	 * Fetches a metadata source for the name.
+	 * 
+	 * @param sourceName The metadata source name.
+	 * @param service
+	 * @return The {@MetadataSource } instance if found, throws exception otherwise.
+	 */
+	public MetadataSource fetchMetadataSourceByName(String sourceName) throws IllegalArgumentException {
```
mks-d: Same here.

api/src/main/java/org/openmrs/module/initializer/api/metadata/MetadataTermMappingsLineProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+		super(headerLine, service);
+	}
+	
+	@Override
+	protected MetadataTermMapping bootstrap(CsvLine line) throws IllegalArgumentException {
+		String uuid = getUuid(line.asLine());
+		String sourceName = line.get(MAPPING_SOURCE, true);
+		String mappingCode = line.get(MAPPING_CODE, true);
+		
+		MetadataTermMapping mapping = service.fetchMetadataTermMapping(sourceName, mappingCode);
```
mks-d: You should also try to bootstrap the entity by UUID if one is provided. What if you want to ever edit the mapping code of a existing metadata mapping? The only way will then be to reference it by UUID.

api/src/main/java/org/openmrs/module/initializer/api/metadata/MetadataTermMappingsLineProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+			}
+		}
+		mapping.setRetired(getVoidOrRetire(line.asLine()));
+		
+		return mapping;
+	}
+	
+	@Override
+	protected MetadataTermMapping fill(MetadataTermMapping mapping, CsvLine line) throws IllegalArgumentException {
+		MetadataSource source = service.fetchMetadataSourceByName(line.get(MAPPING_SOURCE));
```
mks-d: Surely the MAPPING_SOURCE column is also mandatory.

api/src/main/java/org/openmrs/module/initializer/api/loaders/MetadataMappingsLoader.java
```diff
@@ -0,0 +1,35 @@
...
+import org.openmrs.module.initializer.api.metadata.MetadataTermMappingCsvParser;
+import org.openmrs.module.initializer.api.metadata.MetadataMappingServiceWrapper;
+import org.openmrs.module.metadatamapping.api.MetadataMappingService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+import java.io.IOException;
+import java.io.InputStream;
+
+@Component
```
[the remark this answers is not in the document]
mks-d: Did you look at the MDS loader? You will not use `@Component` here.

api/src/test/java/org/openmrs/module/initializer/api/MetadataMappingsLoaderIntegrationTest.java
```diff
@@ -0,0 +1,111 @@
...
+		metadataMappingsLoader.load();
+		
+		// Verify created
+		mtm = service.getMetadataTermMapping(ms, "emr.atFacilityVisitType");
+		Assert.assertNotNull(mtm);
+		Assert.assertEquals("org.openmrs.VisitType", mtm.getMetadataClass());
+		Assert.assertEquals("7b0f5697-27e3-40c4-8bae-f4049abfb4ed", mtm.getMetadataUuid());
+		
+		// Verify edited using source and code
+		mtm = service.getMetadataTermMapping(ms, "emr.primaryIdentifierType");
```
mks-d: If you decide to fetch it by source + code then you should assert the UUID.
I know it was bootstrapped by UUID, but the current test doesn't validate that. I would suggest to assert:
1. all properties it was not bootstrapped with, **and**
1. all properties it was not fetched with in the test.

But anyway here is what I think should be done. You should basically fetch in your tests through the same ID used to bootstrap in the code, that's what we really want to validate.

api/src/main/java/org/openmrs/module/initializer/Domain.java
```diff
@@ -6,16 +6,17 @@
...
-	LOCATIONS(6, "locations"),
-	CONCEPTS(7, "concepts"),
-	PROGRAMS(8, "programs"),
-	PROGRAM_WORKFLOWS(9, "programworkflows"),
-	PROGRAM_WORKFLOW_STATES(10, "programworkflowstates"),
-	PERSON_ATTRIBUTE_TYPES(11, "personattributetypes"),
-	IDENTIFIER_SOURCES(12, "idgen"),
-	DRUGS(13, "drugs"),
-	ORDER_FREQUENCIES(14, "orderfrequencies");
+	METADATA_MAPPINGS(5, "metadatamappings"),
```
mks-d: I reckon it could become the loader with order=2, before MDS. The rule of thumb is that domains should be loaded as early as possible.

Ruhanga: I have updated the PR @mks-d .  :-)

PR #26, opened 2019-11-28
OPENMRS-89: Added ability to process Bahmni appointment service objects

rubailly:
Note that on top of the few properties of an appointment service, it can also have multiple ServiceWeeklyAvailability and AppointmentServiceType linked to it. See [here](https://github.com/Bahmni/openmrs-module-appointments/blob/master/api/src/main/java/org/openmrs/module/appointments/model/AppointmentServiceDefinition.java#L26).


rubailly:
Since AppointmentServiceType and ServiceWeeklyAvailability require an AppointmentServiceDefinition to be created, we can't create those at the same time when locating an AppointmentServiceDefinition CSV file. They'll have to be handled separately.

PR #29, opened 2019-12-24
OPENMRS-94: Add Concept Attribute Processor

chinzou:
https://mekomsolutions.atlassian.net/browse/OPENMRS-94

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
@@ -1,0 +1,70 @@
+package org.openmrs.module.initializer.api.c;
+
+import org.openmrs.Concept;
+import org.openmrs.ConceptAttribute;
+import org.openmrs.ConceptAttributeType;
+import org.openmrs.api.ConceptService;
+import org.openmrs.api.context.Context;
+import org.openmrs.customdatatype.CustomDatatypeUtil;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.openmrs.module.initializer.api.utils.Utils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Component("initializer.conceptAttributeLineProcessor")
+public class ConceptAttributeLineProcessor extends ConceptLineProcessor {
+	
+	@Autowired
+	public ConceptAttributeLineProcessor(@Qualifier("conceptService") ConceptService conceptService) {
+		super(conceptService);
+	}
+	
+	@Override
+	public Concept fill(Concept concept, CsvLine line) throws IllegalArgumentException {
+		
+		Map<String, String> conceptAttributesMap = Utils.getAttributeTypeHeadersMap(line.getHeaderLine());
+		
+		List<ConceptAttributeType> conceptAttributeTypes = conceptService.getAllConceptAttributeTypes();
+		
+		// Filtering out non existing Attribute types
+		Map<String, String> filteredAttributeTypes = conceptAttributesMap.entrySet().stream().filter(map -> {
+			for (ConceptAttributeType ca : conceptAttributeTypes) {
+				if (ca.getName().equals(map.getValue()) || ca.getUuid().equals(map.getValue()))
+					return true;
+			}
+			if (line.get(map.getKey()).isEmpty())
+				return false;
+			throw new IllegalArgumentException("Attribute Type does not exist");
+		}).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+		
+		// Clear Concept attributes before filling
+		Set<ConceptAttribute> cas = concept.getAttributes();
+		cas.clear();
+		
+		// Fill the Concept Attributes Set
+		filteredAttributeTypes.forEach((attHeader, attribute) -> {
+			ConceptAttribute ca = new ConceptAttribute();
+			ConceptAttributeType cat = conceptService.getConceptAttributeTypeByUuid(attribute);
+			if (cat == null) {
+				cat = conceptService.getConceptAttributeTypeByName(attribute);
+			}
+			ca.setAttributeType(cat);
+			ca.setConcept(concept);
+			ca.setCreator(Context.getAuthenticatedUser());
+			ca.setDateCreated(new Date());
+			ca.setValue(CustomDatatypeUtil.getDatatype(ca.getAttributeType()).fromReferenceString(line.get(attHeader)));
+			CustomDatatypeUtil.saveIfDirty(ca);
+			cas.add(ca);
+		});
+		
+		concept.setAttributes(cas);
+		return concept;
+	}
+}
--- a/api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
@@ -5,9 +5,10 @@ import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
-import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.UUID;
 
 import org.apache.commons.lang3.StringUtils;
@@ -44,6 +45,10 @@ import org.openmrs.module.appointments.service.SpecialityService;
 import org.springframework.util.CollectionUtils;
 
 public class Utils {
+	
+	private static String HEADER_ATTRIBUTE = "attribute";
+	
+	private static String ATTRIBUTE_SEPARATOR = "|";
 	
 	/**
 	 * Helps build a {@link ConceptMap} out the usual string inputs.
@@ -441,5 +446,31 @@ public class Utils {
 			}
 		}
 		return false;
+	}
+	
+	/**
+	 * @param headerLine
+	 * @return A Map of Attribute headers and Attribute names or/and UUIDs
+	 * @throws IllegalArgumentException
+	 */
+	public static Map<String, String> getAttributeTypeHeadersMap(String[] headerLine) {
+		
+		Map<String, String> attributeTypes = new HashMap<String, String>();
+		
+		for (String header : headerLine) {
+			String[] parts = StringUtils.split(header, ATTRIBUTE_SEPARATOR);
+			if (parts.length == 2 && parts[0].trim().equals(HEADER_ATTRIBUTE)) {
+				String attributeTypeRef = parts[1].trim();
+				if (attributeTypes.containsValue(attributeTypeRef)) {
+					throw new IllegalArgumentException(
+					        "The CSV header line cannot contain twice the same attribute type: '" + attributeTypeRef + "'");
+				}
+				if (attributeTypeRef.isEmpty()) {
+					throw new IllegalArgumentException("The attribute type name or Uuid cannot be empty.");
+				}
+				attributeTypes.put(header, attributeTypeRef);
+			}
+		}
+		return attributeTypes;
 	}
 }
--- a/api/src/test/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessorTest.java
+++ b/api/src/test/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessorTest.java
@@ -1,0 +1,93 @@
+package org.openmrs.module.initializer.api.c;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.Concept;
+import org.openmrs.ConceptAttribute;
+import org.openmrs.api.ConceptService;
+import org.openmrs.module.initializer.DomainBaseModuleContextSensitiveTest;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+public class ConceptAttributeLineProcessorTest extends DomainBaseModuleContextSensitiveTest {
+	
+	@Autowired
+	@Qualifier("conceptService")
+	private ConceptService cs;
+	
+	private Concept testConcept;
+	
+	@Before
+	public void setup() throws Exception {
+		executeDataSet("testdata/test-concepts.xml");
+		executeDataSet("testdata/test-concepts-attribute.xml");
+		testConcept = cs.getConcept(5501);
+		
+	}
+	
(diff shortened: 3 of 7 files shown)


api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -373,4 +378,30 @@ public static ConceptClass fetchConceptClass(String id, ConceptService cs) {
...
+	/**
+	 * @param headerLine
+	 * @return A Map of Attribute headers and Attribute names or/and UUIDs
+	 * @throws IllegalArgumentException
+	 */
+	public static Map<String, String> getAttributeHeadersMap(String[] headerLine) {
+		
+		Map<String, String> attributes = new HashMap<String, String>();
+		
+		for (String attributeHeader : headerLine) {
```
rbuisson: rename as `header`

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -373,4 +378,30 @@ public static ConceptClass fetchConceptClass(String id, ConceptService cs) {
...
+			String[] parts = org.apache.commons.lang.StringUtils.split(attributeHeader, ATTRIBUTE_SEPARATOR);
+			if (parts.length == 2 && parts[0].trim().equals(HEADER_ATTRIBUTE)) {
+				String attribute = parts[1].trim();
+				if (attributes.containsValue(attribute)) {
+					throw new IllegalArgumentException(
+					        "The CSV header line cannot contains twice the same attribute: '" + attribute + "'");
```
rbuisson: `contain`

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -373,4 +378,30 @@ public static ConceptClass fetchConceptClass(String id, ConceptService cs) {
...
+				}
+				if (attribute.isEmpty()) {
+					throw new IllegalArgumentException("The attribute name or Uuid cannot be empty.");
```
rbuisson: attribute **type**

api/src/test/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessorTest.java
```diff
@@ -0,0 +1,94 @@
...
+		// Setup
+		String[] headerLine = { "Fully Specified Name:en", "attribute|Unavailable" };
+		String[] line = { "Test", "True" };
+		
+		// Replay
+		ConceptAttributeLineProcessor calp = new ConceptAttributeLineProcessor(cs);
+		Concept c = calp.fill(testConcept, new CsvLine(headerLine, line));
+	}
+	
+	@Test()
```
rbuisson: remove `()`

samuelmale:
Hi @chinzou 

Do you want to rebase your working branch against master to resolve the merge conflicts?

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -373,4 +379,30 @@ public static ConceptClass fetchConceptClass(String id, ConceptService cs) {
...
+	 * @throws IllegalArgumentException
+	 */
+	public static Map<String, String> getAttributeHeadersMap(String[] headerLine) {
+		
+		Map<String, String> attributes = new HashMap<String, String>();
+		
+		for (String header : headerLine) {
+			String[] parts = StringUtils.split(header, ATTRIBUTE_SEPARATOR);
+			if (parts.length == 2 && parts[0].trim().equals(HEADER_ATTRIBUTE)) {
+				String attribute = parts[1].trim();
```
samuelmale: Is member `attribute` instead supposed to be `attributeTypeRef` or just `attributeType` ?

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+				return false;
+			throw new IllegalArgumentException("Attribute Type does not exist");
+		}).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+		
+		// Clear Concept attributes before filling
+		Set<ConceptAttribute> cas = concept.getAttributes();
+		cas.clear();
+		
+		// Fill the Concept Attributes Set
+		filteredAttributes.forEach((attHeader, attribute) -> {
```
samuelmale: Again above, `attribute` looks better to me if it was `attributeTypeRef`

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+					return true;
+			}
+			if (line.get(map.getKey()).isEmpty())
+				return false;
+			throw new IllegalArgumentException("Attribute Type does not exist");
+		}).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+		
+		// Clear Concept attributes before filling
+		Set<ConceptAttribute> cas = concept.getAttributes();
+		cas.clear();
```
samuelmale: I'm confused here, I wonder why we need to clear the existing attribute of a concept? I'm I missing something probably?

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+			ca.setAttributeType(cat);
+			ca.setConcept(concept);
+			ca.setCreator(Context.getAuthenticatedUser());
+			ca.setDateCreated(new Date());
+			ca.setValue(CustomDatatypeUtil.getDatatype(ca.getAttributeType()).fromReferenceString(line.get(attHeader)));
+			CustomDatatypeUtil.saveIfDirty(ca);
+			cas.add(ca);
+		});
+		
+		concept.setAttributes(cas);
```
samuelmale: Again here, I think all you have been doing within this overriden `fill(...)` function is actually is supposed to be done from the `bootstrap(..)` function according to the Initializer workflow see: https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/api/src/main/java/org/openmrs/module/initializer/api/BaseLineProcessor.java#L61-L68

So I would do something like:

```java
 
        @Override
	protected Concept bootstrap(CsvLine line) throws IllegalArgumentException {
		Concept concept = super.bootstrap(line);
                 // play around with the bootstrapped concept instance and add necessary attributes
                 // Bootstrap attributes
                 concept.getAttributes().addAll(attributes);
                 return concept;
	}
   ```

**Comment by samuelmale** [comment]
Thanks for working on this @chinzou. Overall, it looks good to me. However, I think we should have at-least a Spring integration test. This could be within any of the Concept loader tests, lets consider an incidence where we are processing a csv line with a concept attribute.

I left a few minor comments.

File: api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java (hunk 8 above)
Comment (samuelmale) [reply api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java]: Not sure if the above would cause a hit on Hibernate Persistence issues though.

Comment (samuelmale) [reply api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java]: I think I'm wrong as per the above, I'm sorry @chinzou ..

File: api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java (hunk 5 above)
Comment (chinzou) [reply api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java]: Thanks @samuelmale , `attributeTypeRef` sounds better since this is an attributeType Name or Uuid

File: api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java (hunk 7 above)
Comment (chinzou) [reply api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java]: Since concept attributes are metadata that we provide via Iniz, we should make sure that the assigned attributes are the ones in the CSV file.
In case we update a concept attributes in our CSVs it's better to clear the existing values before setting the new values

File: api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java (hunk 8 above)
Comment (chinzou) [reply api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java]: According to your provided link, the `bootstrap(..)` function is used to fetch a concept using (Uuid, name, ...) if not found create a new instance and then fill it with `fill(..)`.
I think by checking this package https://github.com/mekomsolutions/openmrs-module-initializer/tree/master/api/src/main/java/org/openmrs/module/initializer/api/c , we can see that all the classes that extends from `ConceptLineProcessor` are overriding `fill(..)` to add data to the concept.
why you think it's better if we override `bootstrap(..)`?

**Comment by chinzou** [comment]
Thanks @samuelmale for your review ^^, you can find here the Tests for ConceptAttributeLineProcessor: https://github.com/mekomsolutions/openmrs-module-initializer/pull/29/files#diff-54920c534372eef91e03d9559d1084a7

File: api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java (hunk 7 above)
Comment (samuelmale) [reply api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java]: What would happen incases were a concept already had existing attributes? Wouldn't this detach it's existing concepts?

**Comment by samuelmale** [comment]
> Thanks @samuelmale for your review ^^, you can find here the Tests for ConceptAttributeLineProcessor: https://github.com/mekomsolutions/openmrs-module-initializer/pull/29/files#diff-54920c534372eef91e03d9559d1084a7

Sure, though I meant an integration Spring test. You could add a test to the [ConceptsLoaderIntegrationTest](https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/api/src/test/java/org/openmrs/module/initializer/api/ConceptsLoaderIntegrationTest.java) that shows concept attributes as part of the test csv files.

**Comment by samuelmale** [comment]
Btw, I'm interested in this cause I'm working on similar work. Adding support for LocationAttribute so we could have a mutual design on spiking on this see: https://github.com/mekomsolutions/openmrs-module-initializer/pull/30

File: api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java [hunk 9]
```diff
@@ -0,0 +1,70 @@
...
+		Set<ConceptAttribute> cas = concept.getAttributes();
+		cas.clear();
+		
+		// Fill the Concept Attributes Set
+		filteredAttributeTypes.forEach((attHeader, attribute) -> {
+			ConceptAttribute ca = new ConceptAttribute();
+			ConceptAttributeType cat = conceptService.getConceptAttributeTypeByUuid(attribute);
+			if (cat == null) {
+				cat = conceptService.getConceptAttributeTypeByName(attribute);
+			}
```
mks-d: Isn't there already a util function to fetch an attribute type by all possible means? If not then we should add one.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+					return true;
+			}
+			if (line.get(map.getKey()).isEmpty())
+				return false;
+			throw new IllegalArgumentException("Attribute Type does not exist");
+		}).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+		
+		// Clear Concept attributes before filling
+		Set<ConceptAttribute> cas = concept.getAttributes();
+		cas.clear();
```
mks-d: Why not just?
```java
concept.setAttributes(null);
```
And then adding them one by one?

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+				if (ca.getName().equals(map.getValue()) || ca.getUuid().equals(map.getValue()))
+					return true;
+			}
+			if (line.get(map.getKey()).isEmpty())
+				return false;
+			throw new IllegalArgumentException("Attribute Type does not exist");
+		}).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+		
+		// Clear Concept attributes before filling
+		Set<ConceptAttribute> cas = concept.getAttributes();
```
mks-d: Name it `attributes`.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+		cas.clear();
+		
+		// Fill the Concept Attributes Set
+		filteredAttributeTypes.forEach((attHeader, attribute) -> {
+			ConceptAttribute ca = new ConceptAttribute();
```
mks-d: Name it `attribute`.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+			ConceptAttributeType cat = conceptService.getConceptAttributeTypeByUuid(attribute);
```
mks-d: Name it `attType`.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+			if (cat == null) {
+				cat = conceptService.getConceptAttributeTypeByName(attribute);
+			}
+			ca.setAttributeType(cat);
+			ca.setConcept(concept);
+			ca.setCreator(Context.getAuthenticatedUser());
+			ca.setDateCreated(new Date());
+			ca.setValue(CustomDatatypeUtil.getDatatype(ca.getAttributeType()).fromReferenceString(line.get(attHeader)));
+			CustomDatatypeUtil.saveIfDirty(ca);
+			cas.add(ca);
```
mks-d: Are you not supposed to simply do?
```java
concept.addAttribute(ca);
```

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptAttributeLineProcessor.java
mks-d: @samuelmale it is paramount with Iniz that CSV lines are [WYSIWYG](https://en.wikipedia.org/wiki/WYSIWYG). Iniz should always make extensive use of **all** available setters on an entity.

In short it should clear it and then refill it as per the CSV line.

mks-d:
@chinzou closing this one because of https://github.com/mekomsolutions/openmrs-module-initializer/pull/56

PR #42, opened 2020-01-28
CSV parser to satisfy the RFC4180 standard.

brandones:
The default CSV Parser used by OpenCSV does not satisfy the CSV standard, RFC4180. It handles escape characters incorrectly. Because of this, OpenCSV also offers a CSV parser called RFC4180Parser. This is the correct parser to use.

PR #43, opened 2020-01-28
Support for PatientIdentifierType.

brandones:


api/src/test/resources/testAppDataDir/configuration/patientidentifiertypes/pit.csv
```diff
@@ -0,0 +1,5 @@
+"Uuid","Void/Retire","Name","Description","Required","Format","Format description","Validator","Location behavior","Uniqueness behavior"
+"73f4f1d6-6086-41d5-a0f1-6d688a4b10af",,"Hospital ID","The patient ID for the hospital. You know the one",TRUE,"H-\\d\\d\\d\\d-\\d","Must have format H-####-C where C is the Luhn check digit","org.openmrs.patient.impl.LuhnIdentifierValidator","NOT_USED","UNIQUE"
+,TRUE,"Old Identification Number",,,,,,,
+,,"Test National ID No","Changing all the fields",TRUE,"\\a\\a\\a\\d\\d\\d\\d","Three letters and four numbers","org.openmrs.patient.impl.LuhnIdentifierValidator","REQUIRED","LOCATION"
```
brandones: Backslashes need to be escaped until https://github.com/mekomsolutions/openmrs-module-initializer/pull/42 is in. Backslashes are not escape characters in the CSV standard.

api/src/main/java/org/openmrs/module/initializer/Domain.java
```diff
@@ -2,19 +2,24 @@
...
-	                                "programs"), PROGRAM_WORKFLOWS(13, "programworkflows"), PROGRAM_WORKFLOW_STATES(14,
-	                                        "programworkflowstates"), PERSON_ATTRIBUTE_TYPES(15,
-	                                                "personattributetypes"), IDENTIFIER_SOURCES(16, "idgen"), DRUGS(17,
-	                                                        "drugs"), ORDER_FREQUENCIES(18, "orderfrequencies"), ORDER_TYPES(
-	                                                                19, "ordertypes"), APPOINTMENTS_SPECIALITIES(20,
-	                                                                        "appointmentsspecialities"), APPOINTMENTS_SERVICES_DEFINITIONS(
-	                                                                                21,
-	                                                                                "appointmentsservicesdefinitions"), DATAFILTER_MAPPINGS(
-	                                                                                        22, "datafiltermappings");
+	JSON_KEY_VALUES(1, "jsonkeyvalues"), MDS(2, "metadatasharing"), PATIENT_IDENTIFIER_TYPES(3,
```
mseaton: Any particular reason you added this in as the 3rd item in the domain?  I think it's fine here, as it doesn't have any dependencies, but might be more logically grouped right before "IDENTIFIER_SOURCES".

Not on you, but the formatting of this class needs a little love.

brandones: Yeah, it seems like one of the more common Metadata Mappings that gets set up is `emr.primaryIdentifierType`. MDM should probably be refactored to go at the end, since it seems like it would reference all of the other domains. But I'd do that in a separate PR.

PR #51, opened 2020-03-17
Renaming MetadataMappings* to MetadataTermMappings*

Ruhanga:


api/src/main/java/org/openmrs/module/initializer/Domain.java
```diff
@@ -5,7 +5,7 @@
 	JSON_KEY_VALUES(1, "jsonkeyvalues"),
 	MDS(2, "metadatasharing"),
 	PATIENT_IDENTIFIER_TYPES(3, "patientidentifiertypes"),
-	METADATA_MAPPINGS(4, "metadatamappings"),
+	METADATA_TERM_MAPPINGS(4, "metadatatermmappings"),
```
mks-d: Reorder that one quite at the bottom, like metadata sets and sets members.

Ruhanga: @mks-d, this is now rectified.

PR #60, opened 2020-04-14
PatientIdentifierType#required better documented.

rbuisson:
As per Slack conversation:
https://openmrs.slack.com/archives/CPC20CBFH/p1586850691014900£

> One small thing I’ve noticed though: the headers in the readme are missing the Required field, and it happens that this field is the only one that is mandatory (other than the name) in order to successfully save the line.
> So 2 ways: either to specify this in the Readme, or we consider a default value when the field value is not provided.

Went for defaulting to `FALSE` when the field value is not provided.

### DIFF
--- a/readme/pit.md
+++ b/readme/pit.md
@@ -7,10 +7,10 @@ patientidentifiertypes/
 ```
 There is currently only one format for the patient identifier type CSV line, here are the possible headers:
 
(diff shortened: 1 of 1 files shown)


rbuisson:
Well if that's what the Iniz usually does then I am OK with that.
Though note that the header is not strictly missing, but instead the field does not have a value set for this header, if that makes any difference

rbuisson:
As discussed, Iniz should not set default values.
Simply updating the README to inform that the field is mandatory.
@mks-d , could review?

Issue #63, opened 2020-04-24
Initializer doesn't support non-Java locales


mogoodrich:
When loading in messages.properties files, Iniz parses the locales for each of the files and validates that each locale is in the "available locales" list:

https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java#L177

This test confirms that the locale in the list of Java support locales:

https://www.oracle.com/java/technologies/javase/jdk8-jre8-suported-locales.html

I'd like to submit that this is to restrictive, and we should be able to drop this check entirely.  In the PIH-EMR we use the Haitian Kreyol locale ("ht") and define a "messages_ht.properties" file.  Even though this is evidently not a "supported" locale in Java, it has worked fine with our OpenMRS implementation, but, unfortunately, fails with Iniz because it is flagged as invalid.

mogoodrich:
PR here: https://github.com/mekomsolutions/openmrs-module-initializer/pull/64

mks-d:
Thanks @mogoodrich .

mogoodrich:
Thanks @mks-d !

PR #64, opened 2020-04-24
[#63] Relax check for JRE-installed locales.

mogoodrich:


api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
```diff
@@ -174,9 +174,7 @@ public Locale getLocaleFromFileBaseName(String baseName) throws IllegalArgumentE
 				continue;
 			}
 			
-			if (LocaleUtils.isAvailableLocale(locale)) {
-				return locale;
-			}
+			return locale;
```
mks-d, line 177: @mogoodrich can't we override/wrap `LocaleUtils.isAvailableLocale(Locale)` and add missing exceptions such as `ht` or `ht_HT`?

mogoodrich: We could, but that doesn't seem to be the "correct" thing to do.  What I don't 100% understand is the relevance of a "available" locale.  

Note that I just tested a "String[] locales = Locale.getISOLanguages()" does return "ht" in it's list.

"getAvailableLocales()" just returns the locales "installed in the JRE".  Not sure exactly what this means, but we've successfully used "ht" with OpenMRS for years even though it's not "installed in the JRE", which I think means it just doesn't have the appropriate services to do date formatting, etc (https://docs.oracle.com/javase/7/docs/api/java/util/spi/LocaleServiceProvider.html)... we could (should?) add our custom versions of these for Haiti at some point, but in the meantime I wouldn't want to fake it into thinking they are installed.

PR #69, opened 2020-06-01
InitializerMessageSource consistent with Spring MessageSources.

ibacher:
I was working with PIH to help them track down an issue where single-quotes in some messages were disappearing even though they had been around for quite a while. The change introduced in their environment was that these values were being loaded by the `InitializerMessageSource`. That makes perfect sense, since the `InitializerMessageSource` loads every property as a [`MessageFormat`](https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html) object and `MessageFormat` objects treat single-quotes as a special character requiring them to be escaped.

The issue is that this hadn't happened prior to using Initializer and there aren't any hacks in to OpenMRS codebase to escape single-quotes. What appears to have happened is that Spring has introduced a certain API quirk (you can read about it [in the Javadocs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/AbstractMessageSource.html)) where when a `MessageSource` is asked for message with no other arguments, it returns the raw property string, but when a `MessageSource` is asked for a message *with* arguments, it uses a `MessageFormat` instance to format the message. However, although this implementation quirk is true of all Spring-implemented `MessageSource` instances, it is not enforced by the [`AbstractMessageSource`](https://github.com/spring-projects/spring-framework/blob/2e7ed915cd8bdac0e92e10717ca48c03a1304a8c/spring-context/src/main/java/org/springframework/context/support/AbstractMessageSource.java) class, which merely requires subclasses to implement the function that takes a code and locale and returns a `MessageFormat` object.

This PR changes the `InitializerMessageSource` so that its behaviour is consistent with Spring implemented `MessageSource` instances. That means that if no arguments are needed for a message, the message doesn't need escaping but when arguments are needed, the message does need escaping.

```
country=Côte d'Ivoire
country.prefixed=Côte d''Ivoire {0}
```

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
+++ b/api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
@@ -324,6 +324,22 @@ public class InitializerMessageSource extends AbstractMessageSource implements M
 		if (showMessageCode) {
 			return new MessageFormat(code);
 		}
+		
+		String message = resolveCodeWithoutArguments(code, locale);
+		
+		if (message != null) {
+			return new MessageFormat(message);
+		}
+		
+		return null;
+	}
+	
+	@Override
+	protected String resolveCodeWithoutArguments(String code, Locale locale) {
+		if (showMessageCode) {
+			return code;
+		}
+		
 		PresentationMessage pm = getPresentation(code, locale); // Check exact match
 		if (pm == null) {
 			if (locale.getVariant() != null) {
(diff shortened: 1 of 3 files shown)


mogoodrich:
Thanks @ibacher !  fyi @mks-d ....

mks-d:
Thanks a bunch @ibacher!

mogoodrich:
Thanks @ibacher  and @mks-d !

PR #71, opened 2020-06-15
Added test case to validate message with apostrophe and arguments

rrameshbtech:
Based on my understand of below PR, I have added one test to expect a message with apostrophe & arguments. This message has single apostrophe. This test case fails.

https://github.com/mekomsolutions/openmrs-module-initializer/pull/69

@ibacher , @mks-d is this test case expected to fail?

### DIFF
--- a/api/src/test/resources/testAppDataDir/configuration/messageproperties/metadata_fr.properties
+++ b/api/src/test/resources/testAppDataDir/configuration/messageproperties/metadata_fr.properties
@@ -1,3 +1,4 @@
 metadata.healthcenter=Clinique
 metadata.healthcenter.description=Ceci est la description d'une clinique.
 metadata.healthcenter.description.named=Ceci est la description de la clinique {0}.
+metadata.healthcenter.details=Voici l'exemple d'apostrophe {0}.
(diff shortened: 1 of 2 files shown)


api/src/test/resources/testAppDataDir/configuration/messageproperties/metadata_fr.properties
```diff
@@ -1,3 +1,4 @@
 metadata.healthcenter=Clinique
 metadata.healthcenter.description=Ceci est la description d'une clinique.
 metadata.healthcenter.description.named=Ceci est la description de la clinique {0}.
+metadata.healthcenter.details=Voici l'exemple d'apostrophe {0}.
```
mks-d, line 4: @rrameshbtech so this is showing that single quote + argument works, thereby contradicting @ibacher's remark [here](https://github.com/mekomsolutions/openmrs-module-initializer/pull/69#issue-426185375), correct?

rrameshbtech: Sorry. I missed. This is failing test case.

mks-d: @rrameshbtech if the test fails then that's in line with @ibacher's conclusion.

Does this make the test pass?
```java
metadata.healthcenter.details=Voici l''exemple d'apostrophe {0}.
```
So with the double `''`.

ibacher: @rrameshbtech As I remarked in the link @mks-d pointed out, this is odd, but (it seems) expected behaviour. See [the Javadoc for `AbstractMessageSource#resolveCodeWithoutArguments(String, Locale)`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/AbstractMessageSource.html#resolveCodeWithoutArguments-java.lang.String-java.util.Locale-).

Issue #65, opened 2020-05-07
Support for 'RelationshipType'


mks-d:
Some further reading, credits to @brandones:
* OpenMRS Talk: [“Preferred” relationship types](https://talk.openmrs.org/t/preferred-relationship-types/34418?u=mksd)

PR #72, opened 2020-06-15
[#65] Support for RelationshipType.

samuelmale:
Issue: https://github.com/mekomsolutions/openmrs-module-initializer/issues/65


api/src/main/java/org/openmrs/module/initializer/api/relationships/types/RelationshipTypesCsvParser.java
```diff
@@ -0,0 +1,50 @@
...
+	@Override
+	public RelationshipType bootstrap(CsvLine line) throws IllegalArgumentException {
+		RelationshipType relationshipType = null;
+		String uuid = line.getUuid();
+		if (StringUtils.isNotBlank(uuid)) {
+			relationshipType = ps.getRelationshipTypeByUuid(uuid);
+		}
+		if (relationshipType == null) {
+			relationshipType = new RelationshipType();
+			relationshipType.setUuid(uuid);
```
Ruhanga, line 35: Thanks for the nice work on this @samuelmale. Just a little comment here. It could happen that the `uuid` is an empty string in which case you'd simply leave that field untouched.

Ruhanga: @samuelmale, you may also add documentation and tests for this as well.

samuelmale: > Thanks for the nice work on this @samuelmale. Just a little comment here. It could happen that the `uuid` is an empty string in which case you'd simply leave that field untouched.

@Ruhanga We normally leave validation to the dedicated Validators in the Openmrs API or service layers.

samuelmale: > @samuelmale, you may also add documentation and tests for this as well.

I hardly understand, for what?

Ruhanga: >@Ruhanga We normally leave validation to the dedicated Validators in the Openmrs API or service layers.
With a `new` instance, a `uuid` is generated so you don't want to replace that with an empty string that'd mess with the validations, if you understand what I mean.

Ruhanga: > hardly understand, for what?

The `uuid` field needs to be documented like all other fields are. Not everyone knows why it exists, how it is used, whether mandatory or not. Also this should be fully covered in tests cases. Does this makes sense?

samuelmale: >With a new instance, a uuid is generated so you don't want to replace that with an empty string that'd mess with the validations, if you understand what I mean.

Nice catch @Ruhanga

samuelmale: Sure I think we need to document all those generic fields with the rest of domains in Iniz 😃

Ruhanga: Thanks @samuelmale. You can create a separate issue to address this documentation improvement on the other domains whilst you address that part relevant to this PR here. @mks-d any thought on this?

mks-d: @Ruhanga @samuelmale did you see [this](https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/csv_conventions.md#header-uuid-mandatory)?

There's an handful of headers that are pretty much always there, and they are documented there.

mks-d: >It could happen that the uuid is an empty string in which case you'd simply leave that field untouched.

@samuelmale indeed, like it's done in virtually all domains, like [here](https://github.com/mekomsolutions/openmrs-module-initializer/blob/0ac0eb9ceedbb1d950f4d423f71ad16ae4006978/api/src/main/java/org/openmrs/module/initializer/api/loc/LocationsCsvParser.java#L48-L51) for locations for example.

Ruhanga: >did you see this?

Thanks for that pointer @mks-d. The documentation already exists actually. We can then simply implement this for this PR @samuelmale.

api/src/main/java/org/openmrs/module/initializer/api/relationships/types/RelationshipTypesCsvParser.java
```diff
@@ -32,7 +32,9 @@ public RelationshipType bootstrap(CsvLine line) throws IllegalArgumentException
 		}
 		if (relationshipType == null) {
 			relationshipType = new RelationshipType();
-			relationshipType.setUuid(uuid);
+			if (StringUtils.isNotBlank(uuid)) {
+				relationshipType.setUuid(uuid);
+			}
```
Ruhanga, line 37: @samuelmale, could this not be covered in a test?

brandones:
Suggest closing in favor of https://github.com/mekomsolutions/openmrs-module-initializer/pull/129

Issue #44, opened 2020-02-20
Add full support for installing location tags


mogoodrich:
Currently Iniz does not have full support for creating location tags outside of a location… currently, you just specific location tags by name “inline” in the locations.csv.

When saving a location, Iniz looks up each of these location tags by name, and creates them as needed.  

For this ticket, we want to make location tags first-class citizens.  

A first-pass at a design:

* Should support a locationtags.csv that allows specifying the location tag uuid, name and description

* The Locations loader should be updated so that it allows referring to tags by both uuid and name

* The Locations loader should also be updated so that if a “locationtags.csv” is present, it doesn’t “hot create” location tags as-needed as it currently does, rather if a location tag referenced in location.csv is not found, it should through a validation error; if no locationtags.csv file is present, it should continue to work as it currently does.

@mks-d thoughts on this?

fyi @mseaton  @brandones

mseaton:
I 100% agree with making locationtags a first-class citizen, and for it to load ahead of the locations. Based on how other domains work (eg. Concept Class is a separate domain from Concept), it would seem that Location Tag should be a separate domain that is ordered ahead of Location.  This would be straightforward to implement.

Regarding how to handle this in relation to the current behavior with the "Tags" column in the Location domain, my vote would be to have this column retain it's current behavior for backwards compatibility.  Then, we would add support for additional columns in the same manner that Location Attributes are supported.  There would be columns with headers named (for example):

`Tag|Login Location`

With a column value that is either "true" or "false/blank".

This would only support looking up tags by name or uuid (based on the column value), it would not support hot-creating tags on demand.

This would lead to more intuitive Location CSVs, which - instead of having a bloated column value with lots of delimited tag names, simply has a bunch of extra columns with "true" or "" values indicating whether or not that Location has that tag.

This would also allow easier management of the metadata - one can easily take this CSV in Excel and sort it or filter it to show all of the Login Locations or all of the Visit Locations, etc.

mks-d:
@mogoodrich 👍 sure, let's add that new domain.

@mseaton I like your idea, and obviously it is even better if you see Excel value for it.

The current "hot loading" was not properly thought through, and this is something that is overdue. Either to conclude that hot loading should be made smarter and more configurable, or should be deprecated (however, think of `ConceptReferenceTerm` for instance...). This is a discussion that can happen elsewhere. First of all I need to make an inventory of where exactly it happens, at worst in one or two other instances than location tags.

[...]

PR #73, opened 2020-07-01
[#44] Location domain to support for 'Tag|' headers.

brandones:
addresses #44

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
@@ -1,6 +1,8 @@
 package org.openmrs.module.initializer.api.loc;
 
+import java.util.Arrays;
 import java.util.HashSet;
+import java.util.function.Consumer;
 
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.Location;
@@ -20,7 +22,9 @@ public class LocationLineProcessor extends BaseLineProcessor<Location> {
 	protected static String HEADER_PARENT = "parent";
 	
 	protected static String HEADER_TAGS = "tags";
-	
+
+	public static final String HEADER_TAG_PREFIX = "tag|";
+
 	protected static String HEADER_CITY_VILLAGE = "city/village";
 	
 	protected static String HEADER_COUNTY_DISTRICT = "county/district";
@@ -45,13 +49,13 @@ public class LocationLineProcessor extends BaseLineProcessor<Location> {
 	
 	private LocationService locationService;
 	
-	private LocationTagListParser listParser;
-	
+	private LocationTagListParser tagListParser;
+
 	@Autowired
 	public LocationLineProcessor(@Qualifier("locationService") LocationService locationService,
 	    LocationTagListParser listParser) {
 		this.locationService = locationService;
-		this.listParser = listParser;
+		this.tagListParser = listParser;
 	}
 	
 	@Override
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
```diff
@@ -86,4 +91,24 @@ public Location fill(Location loc, CsvLine line) throws IllegalArgumentException
...
+						+ "' exists for header '" + HEADER_TAG_PREFIX + tagName);
+			}
+			location.addTag(tag);
+		};
+
+		// process the value for each tag header
+		Arrays
+				.stream(line.getHeaderLine())
+				.filter(h -> StringUtils.startsWithIgnoreCase(h, HEADER_TAG_PREFIX))
+				.filter(h -> Boolean.TRUE.equals(line.getBool(h)))
```
mks-d, line 110: @brandones @mogoodrich does that mean that the `false` value is never processed? I would expect the `false` value to mean that the location does **not** carry the tag. In other words it should be actively removed if the location was tagged with it.

Sorry if I missed designs discussions. The README doesn't say any more than that either.

mogoodrich: Defer to @brandones  on what the code does, but I would agree with you @mks-d  on that is the behavior I'd expect/want.

brandones: All tags are automatically cleared at the beginning, see line 73, `loc.setTags(null);`

api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
```diff
@@ -86,4 +91,24 @@ public Location fill(Location loc, CsvLine line) throws IllegalArgumentException
 		
 		return loc;
 	}
+
+	private void setLocationTagsFromPrefixHeaders(Location location, CsvLine line) {
```
mks-d, line 95: @brandones have you considered introducing an ad-hoc line processor altogether?

brandones: I did consider it. When I looked at the existing file names,
Screenshot from 2020-07-02 10-59-56
I didn't see a way to make it obvious that the line processor corresponds to the Location domain rather than the Location Tags domain. But maybe I should have broken up the directories by domain, into `loc`, `loctag`, and `loctagmap` directories. Though perhaps then we would also want to break up the `c` directory.

readme/loc.md
```diff
@@ -7,32 +7,49 @@ locations/
...
+###### `Tag|` headers
+Tags can be assigned in true/false columns under headers starting with `Tag|`
+followed by the location tag _name_, e.g. `Tag|Login Location`.
+Tags assigned in this way will _not_ be created automatically. They can be
+created using the [locationtags](./loctags.md) domain.
 
 ###### Header `Tags`
-The list of location tags. Location tag _names_ should be provided. Those names can point to either existing location tags, or if no existing tag is found a new tag with the specified name will be created.
+A comma-delimited list of location tags specified by _name_. These can be the names 
+of existing location tags, or new tags to be created ad-hoc.
```
mks-d, line 35: Sorry one last thing. We need to document what happens when the two approaches appear to collide, so when one does something like that:
| Tags           | Tag\|Login Location |
|----------------|--------------------|
| Login Location | false              |

I think the outcome is that the tag will be set since:
1. Tags are first cleared.
1. The `Tags` header is processed first, thereby setting the tag to the location.
1. `Tag\|Login Location` set to `false` has not further effect.

Maybe we can even have a test that asserts that exact case?

brandones: I added a note to the README. That case is already covered in the tests—`locations.csv` has both a `Tags` header and `Tag|` headers. Using "only" the `Tags` header is tested in this context, which means the `Tag|` header is blank/false.

Though I realize now, that means that there's no test for *only* the `Tags` header or only the `Tag|` headers. Should I add some, or is that more testing than is worthwhile?

brandones: Ping @mks-d ?

PR #78, opened 2020-10-01
Support for a LocationTagMap domain

brandones:
The final part of #44


brandones:
@mks-d

brandones:
@mks-d Actually, at PIH we developed another mechanism for attaching tags to locations, so we aren't planning to use this. Should we just close this, or do you think this would be a good thing to have in Iniz?

mks-d:
@brandones no I'll review it ahead of the 2.1.0 release. Since the work is done I'd say let's keep it.

Just curious as to what other mechanism you have adopted, maybe you can provide details on Slack?

brandones:
We have a JSON-based "PIH Config." [Here](https://github.com/PIH/openmrs-config-ces/blob/master/configuration/pih/pih-config-mexico.json)'s the one we use in Mexico, for example. We've added support for a config element named "[locationTags](https://github.com/PIH/openmrs-config-ces/blob/fb6b87457c62b3482f1935b4db7fdb29ff55c739/configuration/pih/pih-config-mexico.json#L41)", which associates locations with tags.

Right now it is very cumbersome to work with, because we haven't adopted the Iniz Location CSV format yet—we're still creating locations using MDD. But soon we'll have locations coming from CSV, with most of their tags set using the `Tag|Xyz Location` headers, and just the `Login Location` tag set on a per-site basis by the PIH Config.

brandones:
This seems to have stalled out.

Issue #86, opened 2020-12-03
Enable htmlforms domain to set published and retired, and load from subdirectories


mseaton:
This issue is for making the following enhancements to the newly added htmlforms domain:

* Support configuring the form.published property
* Support configuring the form and htmlform retired property
* Support for organizing htmlform xml files in subdirectories within the htmlforms domain

This will provide the ability to more fully manage forms, and to enable the ability to organize forms based on implementation preferences.

mseaton:
@mks-d - if you could have a look at this and see what you think that would be very helpful.

mks-d:
@mseaton this PR is an attempt to handle the part "should handle subdirectories": https://github.com/mekomsolutions/openmrs-module-initializer/pull/91

However what's happening to the URI to the HTML form though?

mseaton:
@mks-d - that's cool - great to see that subdirectory support enabled broadly!  Not sure what you mean by the URI to HTML form?  What URI?

mks-d:
What I meant is when implementations use HFE UI's `definitionUiResource`. For instance here for the ENTER mode: https://github.com/openmrs/openmrs-module-htmlformentryui/blob/e7b839235797b13f5d859b6cc16bdb3e06af46e5/omod/src/main/java/org/openmrs/module/htmlformentryui/page/controller/htmlform/BaseEnterHtmlFormPageController.java#L62

As you may recall the Ref App encourages the use of such URIs (see [here](https://github.com/openmrs/openmrs-module-referenceapplication/blob/8ac084672ac56643c169341fb879b9942105fe4d/omod/src/main/resources/apps/visitActions_extension.json#L7)):
```url
/openmrs/htmlformentryui/htmlform/enterHtmlFormWithStandardUi.page?...
  &definitionUiResource=referenceapplication:htmlforms/simpleVisitNote.xml
```
When forms are nested within the providing module's OMOD resources then this `definitionUiResource`'s value must reflect that nesting, for instance URIs like:
```url
definitionUiResource=acmemodule:htmlforms/ncd/initialScreening.xml
```
for NCD forms that have been nested inside such folder: `omod/src/main/webapp/resources/htmlforms/ncd`.

----

What's your take in regards to migrating to a config forms that are being referred to with such URIs currently?

mseaton:
Ah, I see.  This is already supported, we added support for this to uiframework a few months ago or so.  You can do something like this now, and things work:

[...]

PR #87, opened 2020-12-03
[#86] htmlforms to support published and retired XML attributes.

mseaton:
…from subdirectories

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/form/HtmlFormsLoader.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/form/HtmlFormsLoader.java
@@ -58,7 +58,7 @@ public class HtmlFormsLoader extends BaseLoader {
 	@Override
 	public void load() {
 		ConfigDirUtil dirUtil = getDirUtil();
-		for (File file : dirUtil.getFilesRecursively("xml")) { // processing all the XML files inside the domain
+		for (File file : dirUtil.getFiles("xml")) { // processing all the XML files inside the domain
 			
 			String fileName = dirUtil.getFileName(file.getPath());
 			String checksum = dirUtil.getChecksumIfChanged(fileName);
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/openmrs/module/initializer/api/form/HtmlFormsLoader.java
```diff
@@ -55,7 +58,7 @@ protected Domain getDomain() {
 	@Override
 	public void load() {
 		ConfigDirUtil dirUtil = getDirUtil();
-		for (File file : dirUtil.getFiles("xml")) { // processing all the XML files inside the domain
+		for (File file : dirUtil.getFilesRecursively("xml")) { // processing all the XML files inside the domain
 			
 			String fileName = dirUtil.getFileName(file.getPath());
 			String checksum = dirUtil.getChecksumIfChanged(fileName);
```
mks-d, line 64: I think the whole writing and checking of checksums will have to be adapted as well unfortunately, to handle a tree of checksum files rather than just one level. Could you confirm?

mseaton: Good point.  Let's hold off on merging, as I'm working on testing this with our distro right now and will have more information as to how things work out.  As for this particular feature, I think we can likely do without it for now, as I can likely handle this at build time for the distribution, by configuring the assembly.xml file to copy from various source directories into a single htmlforms target directory.  It makes for more maintenance and documentation of our build process and might lead to some head scratching when new directories added to the source don't end up in the distro automatically, but should allows us to meet both the need to organize better in our source structure while not overcomplicating how initializer is configured.

mseaton: @mks-d - I've updated the PR and removed the code that was doing the recursive loading in the htmlforms domain, and am limiting this PR to just adding support for the retired and published properties, which is the main thing I need to add in to be able to use this in rwanda

mks-d (approved):

Issue #84, opened 2020-11-20
Bundle a standalone validator tool


mks-d:
#### Background
See this seed Slack conversation: https://openmrs.slack.com/archives/CPC20CBFH/p1582575953010600

mks-d:
@ibacher about your suggestion [here](https://github.com/mekomsolutions/openmrs-module-initializer/pull/88#pullrequestreview-555505458):
>I wonder if it makes sense to also provide a `--include.<domain>` flag to whitelist the specific files from a domain.

Sure, however what do you think should happen when `--exclude.<domain>` and `--include.<domain>` contradict each other?

ibacher:
@mks-d For consistency, it might make sense to just have the `—include.<domain>` flag with the same `!` to switch to exclude mode. Alternatively, I’d just raise an error if both `—include` and `—exclude` were specified for the same domain without making any attempt to analyse if they contradict one another. Basically, I still think the model of “you’re either specifying a white-list or a black-list” is probably sufficient.

PR #88, opened 2020-12-17
[#84] Added Initializer Validator.

mks-d:


readme/rtprops.md
```diff
@@ -0,0 +1,30 @@
...
+```
+##### Example of exclusion list:
+Just prefix the list with a negative `!` prefix character:
+```java
+initializer.domains='!metadatasharing,privileges,roles'
+```
+This is exactly the same logic as for the Initializer Validator's `--domains` argument.
+### 2) `initializer.exclude.<domain>` _(optional)_
+##### Example
+In the example below all 'concepts' domain files matching the [wildcard patterns](https://docs.oracle.com/cd/E23389_01/doc.11116/e21038/conditions.htm#BABEJGAH) `*diags*` and `*interventions*` will be filtered out:
```
ibacher, line 19: Oracle's Traffic Director actually seems to allow a much richer series of expressions than the `WildcardFilenameFilter`.

```suggestion
In the example below all 'concepts' domain files matching the [wildcard patterns](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FilenameUtils.html#wildcardMatchOnSystem-java.lang.String-java.lang.String-) `*diags*` and `*interventions*` will be filtered out:
```

validator/src/main/java/org/openmrs/module/initializer/validator/ConfigurationTester.java
```diff
@@ -0,0 +1,211 @@
...
+	
+	private String cielFilePath;
+	
+	protected String getAppDataDirPath() {
+		return Paths.get(configDirPath).getParent().toString();
+	}
+	
+	protected void setMariaDB4jProps(Properties props) throws ManagedProcessException {
+		DBConfigurationBuilder config = DBConfigurationBuilder.newBuilder();
+		config.setPort(0); // 0 => autom. detect free port
```
ibacher, line 65: This is actually the default value for `DBConfigurationBuilder`

validator/src/main/java/org/openmrs/module/initializer/validator/ConfigurationTester.java
```diff
@@ -0,0 +1,211 @@
...
+		
+		DB db = DB.newEmbeddedDB(config.build());
```
ibacher, line 67: It seems like it would be a better idea to save the `DBConfiguration` object returned from this and use *that* to get the URL.

validator/src/main/java/org/openmrs/module/initializer/validator/Validator.java
```diff
@@ -0,0 +1,167 @@
...
+		boolean needsReappending = false;
+		if (endsWith(sqlPiece, ";")) {
+			needsReappending = true;
+			sqlPiece = removeEnd(sqlPiece, ";");
+		}
+		return replace(sqlPiece, "\\'", "''") + (needsReappending ? ";" : "");
+	}
+	
+	/**
+	 * Turns an original CIEL SQL dump into a HSQLDB-friendly one.
```
ibacher, line 71: Do we need this if we're now loading this file into a MariaDB instance? I.e., won't MariaDB support the MySQL-specific syntax?

ibacher (commented):
This obviously isn't a full review, but a few comments on some things I noticed.

At a much more high-level note: I like the simple whitelist / blacklist syntax ('!metadatasharing...'), but I wonder if it makes sense to also provide a `--include.<domain>` flag to whitelist the specific files from a domain. I realise that might add to the complexity and may be more appropriate for a follow-on to this PR.

validator/src/main/java/org/openmrs/module/initializer/validator/Validator.java
mks-d: @ibacher well yes... And I think it's because at the end of the day HSQLDB isn't that smart. I suspect it just executes statements one by one behind the scenes and doesn't know how to handle comments. Disappointingly it doesn't _really_ do the equivalent of `mysql openmrs < file.sql` (which is what I was hoping to do via some Java code.)

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -9,38 +9,48 @@
...
+			        .getLogger(InitializerActivator.class.getPackage().getName());
+			logger.addAppender(
+			    Utils.getFileAppender(Paths.get(OpenmrsUtil.getApplicationDataDirectory(), MODULE_ARTIFACT_ID + ".log")));
+			logger.setLevel(Level.WARN);
 		}
 		
-		log.info("Start of initializer module.");
+		Context.getService(InitializerService.class).load(true);
 		
+		log.info("Start of " + MODULE_ARTIFACT_ID + " module.");
```
reagan-meant, line 47: A reminder to move provide two logs 
1. When the module is starting 
2. When the module has started

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -43,10 +48,87 @@
...
+				        "Printing a batch of CSV lines is only supported if they share a common header.");
+			}
+			prevHeader = line.getHeaderLine();
+			data.add(setLineSeparators(line.asLine()));
+		}
+		return System.lineSeparator() + AsciiTable.getTable(prevHeader, data.toArray(new String[0][0]));
+	}
+	
+	public static String pastePrint(List<CsvLine> lines) {
+		String printout = "";
```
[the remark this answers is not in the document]
mks-d: Done, thanks.

validator/src/main/java/org/openmrs/module/initializer/validator/ConfigurationTester.java
mks-d: Thanks, I think I did what you suggested:
```java
DB db = DB.newEmbeddedDB(DBConfigurationBuilder.newBuilder().build());
db.start();
db.createDB("openmrs");
...
props.setProperty(Environment.URL, db.getConfiguration().getURL("openmrs"));
```

PR #92, opened 2021-01-27
[#84] Added unsafe mode for domains loading.

mks-d:


mseaton (approved):
Looks like a good refactor.  I can't say I've gone into enough depth to have been able to fully review all of the intentions, but does look like an improvement overall.  I'm happy as long as the unit tests still pass :)

PR #94, opened 2021-02-09
[#93] Support for AppointmentServiceType.

mks-d:


mks-d (commented):
@rbuisson / @reagan-meant, I'm not asking for an actual review but rather for you guys to pay attention to this: https://github.com/mekomsolutions/openmrs-module-initializer/pull/94/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R152-R153

Could you let me know if there is any other repo than the Haiti config that might be impacted by this change?

mks-d (commented):
@rbuisson not sure if I'd still have the room to do it as part of this PR and before any 2.1.0 release, but making of specialities a first-class citizen domains was certainly way overkill. It should have been something created on the fly, like location tags for locations for instance.

Issue #96, opened 2021-03-24
Support for OCL export ZIP file.


mks-d:

gracepotma:
OMRS Ticket ID: https://issues.openmrs.org/browse/OCLOMRS-950 

Notes page: https://docs.google.com/document/d/1UGFYMRawyfOpVqnxuI1YzrPsdMR0pi8z3_Ze-0ixTGM/edit

Additional Use Case Details: 
- **Enable Concepts to be loaded into Iniz that are managed in OCLOMRS**. Have Iniz use Subscription Module. Looking into loader Iniz using. Goal is to have Iniz understand the OCL .zip file (and the JSON file inside).
- (confirm check-sum is ok) Iniz uses a check-sum to determine whether or not to re-load a file - so whatever we put into Iniz, if it's the same content, then ideally it would have the same check-sum
- **Support offline (and reasonable file size)**. Will support offline loading needs as well / "deploy with USB stick" scenario. (currently 18MB pretty & zipped)
- **Support diff check/review of JSON file** (user friendly to look at). The OCL .zip may not have been originally designed with that use case in mind. JSON needs to be formatted nicely enough, with predictable order, so that diffs/review would be relatively easy. (e.g. if you produce it 3 times, you get the properties in the same order every time). → May need update to OCL Module to support unzipped or zipped file.

ibacher:
Closing as #100 was merged.

PR #100, opened 2021-04-09
[#96] OCLOMRS-950: Support for OCL export ZIP file.

suruchee:
#### Ticket:
https://issues.openmrs.org/browse/OCLOMRS-950


ibacher:
@suruchee The code looks fine to me. A couple of points:

1. This PR is marked as a draft; if you want it to be merged, it should be unmarked as a draft.
2. There's apparently a conflict between your branch and `master`. Could you please update your local branch to resolve the conflict?
3. Please title this PR something descriptive, e.g., "Support for OCL export ZIP file"
4. In the description for the PR it would be helpful to mention the OpenMRS ticket for it.

Once you've taken care of the above points, I'd ask Dmitri or someone with commit access to this repo to review this.

Thanks for your work on this!

suruchee:
Thanks @ibacher  for the review. I have updated with master and changed the PR title and description. 
Thank you so much for the support.

api/src/main/java/org/openmrs/module/initializer/Domain.java
```diff
@@ -32,7 +32,8 @@
 	METADATA_SETS(28, "metadatasets"),
 	METADATA_SET_MEMBERS(29, "metadatasetmembers"),
 	METADATA_TERM_MAPPINGS(30, "metadatatermmappings"),
-	HTML_FORMS(31, "htmlforms");
+	HTML_FORMS(31, "htmlforms"),
+	OCL(32, "openconceptlab");
```
mseaton, line 37: This is very likely not the right spot for this.  The order of the loaders in the Domain need to be set based on dependencies.  As many other domains are dependent on Concept, the OCL domain (assuming we want a separate domain for this) should be adjacent to the concept domain.

api/src/main/java/org/openmrs/module/initializer/Domain.java
```diff
@@ -32,7 +32,8 @@
...
```
mseaton, line 36: It's not 100% clear to me that we want "openconceptlab" to be a separate Domain from Concept.  I'd be interested in other's thoughts as to whether this is preferable, or whether having the Concepts domain capable of saying "if I find a file with .zip extension, then assume it's an OCL zip and try to import it using the OCL module.

mseaton (commented):
Just a few questions / comments on my end.  @mks-d interested to understand your perspective...

api/src/main/java/org/openmrs/module/initializer/Domain.java
mogoodrich: I haven't thought it through deeply, but would it make sense to keep it separate:

* in case OCL ends up packaging more than concepts, or because
*  in case we want to use "zip" packaging for the standard csv model

But just thoughts, for final decision I'm fine with what @mseaton  and @mks-d decide

ibacher: There are two small advantages I see to having a separate OCL domain (over and above what Mark said):

1. It keeps the documentation for both the OCL domain and the Concept domain simpler
2. It keeps the code to load concepts from OCL a bit simpler (because whether or not the loader is loaded at all depends on whether the OCL module is present or not; if we integrate this into the concept domain then we have to add a check for the OCL loader).

The biggest downside I see to having a separate domain is that we lose some flexibility with concept dependencies, i.e., either the OCL domain is loaded before the concept domain or its loaded after the concept domain, meaning that either the OCL domain or the Concept domain can't depend on concepts defined in the other domain to be present when loading. That said, I think OCL packaging is fairly self-contained so loading it before the Concept domain shouldn't have any issues.

I don't really have a strong opinion either way and would be happy to see this implemented as part of the concept domain if that's what @mks-d and @mseaton think is best.

api/src/main/java/org/openmrs/module/initializer/api/loaders/OpenConceptLabLoader.java
```diff
@@ -0,0 +1,40 @@
...
+import org.openmrs.api.context.Context;
+
+@OpenmrsProfile(modules = { "openconceptlab:*" })
+public class OpenConceptLabLoader extends BaseFileLoader {
+	
+	private Importer importer;
+	
+	private Importer getImporter() {
+		if (importer == null) {
+			importer = Context.getRegisteredComponent("openconceptlab.importer", Importer.class);
```
mks-d, line 19: Please autowire rather than using `Context` static methods to get beans. That'll help you get rid of that `if`.

pom.xml
```diff
@@ -187,6 +186,12 @@
             <version>${metadatasharingVersion}</version>
             <scope>provided</scope>
         </dependency>
+           <dependency>
```
mks-d, line 189: There's a misindentation here.

pom.xml
```diff
@@ -61,7 +61,7 @@
         <idgenVersion>4.6.0</idgenVersion>
         <metadatasharingVersion>1.2.2</metadatasharingVersion>
         <metadatamappingVersion>1.3.4</metadatamappingVersion>
-
+		<openconceptlabVersion>1.2.6-SNAPSHOT</openconceptlabVersion>
```
mks-d, line 64: There's a misindentation here.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+		
+		DaemonToken daemonToken = new DaemonToken("openconceptlab");
+		daemonTokens.put(daemonToken.getId(), daemonToken);
+		new OpenConceptLabActivator().setDaemonToken(daemonToken);
+	}
+	
+	@Test
+	public void load_shouldImportOCLPackages() {
+		{
+			Concept c = Context.getConceptService().getConceptByUuid("4421da0d-42d0-410d-8ffd-47ec6f155d8f");
```
mks-d, line 54: Autowire `ConceptService`.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+			field.setAccessible(true);
+			daemonTokens = (Map<String, DaemonToken>) field.get(null);
+		}
+		catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+		
+		DaemonToken daemonToken = new DaemonToken("openconceptlab");
+		daemonTokens.put(daemonToken.getId(), daemonToken);
+		new OpenConceptLabActivator().setDaemonToken(daemonToken);
```
mks-d, line 48: I guess the OCL importer needs this?

api/src/main/java/org/openmrs/module/initializer/Domain.java
mks-d, line 36: The order should be either just before or just after that of concepts.

mks-d, line 36: Maybe
```java
OCL(32, "ocl")
```
?
As for the class and variable names I think I prefer the full wording, and that's what you have done 👍

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+			Assert.assertNull(c);
+		}
+		
+		loader.load();
+		
+		{
+			Concept c = Context.getConceptService().getConceptByUuid("4421da0d-42d0-410d-8ffd-47ec6f155d8f");
+			Assert.assertEquals(2, c.getSetMembers().size());
+			Assert.assertTrue(c.getSet());
+		}
```
mks-d, line 64: I think you should come up with a test .zip file that contains a couple of concepts (and not necessarily a full CIEL). Because then your asserts don't make much sense.

Export a couple of concepts in a .zip and **fully assert** that they have been loaded as expected in OpenMRS.

mks-d (changes requested):

api/src/main/java/org/openmrs/module/initializer/Domain.java
mks-d: I would vote for keeping them separate and load 'ocl' before 'concepts'.
That way 'concepts' could be used to overwrite 'ocl' if ever needed for whatever implementation or site-specific reason.

mseaton: Sounds good to me, thanks @mks-d

suruchee:
Thank you everyone for the review. @mks-d , I have made changes as you have suggested, can you please review it once again.

api/src/main/java/org/openmrs/module/initializer/api/loaders/OpenConceptLabLoader.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.io.File;
+import java.util.zip.ZipFile;
+
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.initializer.Domain;
+import org.openmrs.module.openconceptlab.importer.Importer;
+import org.openmrs.api.context.Context;
+
+@OpenmrsProfile(modules = { "ocl:*" })
```
mks-d, line 11: @suruchee this is most likely going to be
```java
@OpenmrsProfile(modules = { "openconceptlab:1.2.6-SNAPSHOT" })
```

api/src/main/java/org/openmrs/module/initializer/api/loaders/OpenConceptLabLoader.java
mks-d: @suruchee did you see my comment above ↑?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
mks-d: @suruchee did you see my comment above ↑?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+	public void load_shouldImportOCLPackages() {
+		{
+			Concept c = conceptService.getConceptByUuid("5821b91383ac240007c3c0f3");
+			Assert.assertNull(c);
+		}
+		
+		loader.load();
+		
+		{
+			Concept c = conceptService.getConceptByUuid("5821b91383ac240007c3c0f3");
```
mks-d, line 63: @suruchee `5821b91383ac240007c3c0f3` is an illegal UUID.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+			Assert.assertEquals("TETANUS BOOSTER", c.getName());
+			Assert.assertEquals("Procedure", c.getConceptClass());
```
mks-d, line 65: @suruchee surely there's a lot more that you want to assert, all the mappings typically.

You need to demonstrate that you can load a rich set of information about concepts. I would also go beyond just one concept.

Having said that, thanks for trimming down the test .zip.

mks-d (changes requested):
Again, in the same spirit as MDS, you will have to add your new `openconceptlab` dependency for the Validator, see here for MDS: https://github.com/mekomsolutions/openmrs-module-initializer/blob/e21ad7a277c05651f2d1cd24b4516772524ba7f0/validator/pom.xml#L169-L175

I also strongly suggest that you validate your work with the Validator btw.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
ibacher: Yes. The actual loading is done in a daemon thread and the daemon token is fetched from the activator.

suruchee:
@mks-d,I have tried to improve tests, will you please review the changes?

api-2.3/src/test/java/org/openmrs/module/initializer/api/loaders/LoadersOrderTest.java
```diff
@@ -306,6 +309,11 @@ public void getLoaders_shouldBeUnivoquelyOrdered() {
 		count++;
 		Assert.assertThat(loader.getOrder(), greaterThan(previousLoader.getOrder()));
 		
+		previousLoader = loader;
+		loader = openConceptLabLoader;
+		count++;
+		Assert.assertThat(loader.getOrder(), greaterThan(previousLoader.getOrder()));
```
mks-d, line 315: @suruchee this should be move up between those two lines: https://github.com/mekomsolutions/openmrs-module-initializer/pull/100/files#diff-5c58224096e13067818fadbb6a3f8c8ad7b7548c524f8acc80de7f4190f57b7aR220-R222

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
mks-d: @ibacher does it need to involve the real activator or this could be done, see
* [Set a Daemon token in modules context-sensitive tests](https://talk.openmrs.org/t/set-a-daemon-token-in-modules-context-sensitive-tests/32022/5?u=mksd)

Makes me think that I never created the ticket for Core, I wonder if this could be scoped in and then used here?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+	
+	@Test
+	public void load_shouldImportOCLPackages() {
+		Concept c = null;
+		{
+			c = conceptService.getConceptByUuid("db1741f033144371a191649f98454148");
+			Assert.assertEquals("Text", c.getDatatype());
+			Assert.assertFalse(c.getRetired());
+		}
```
mks-d, line 66: @suruchee the reason for the anonymous scope here is so that you can keep redeclaring `Concept c` multiple times. So just do:
```java
{
    Concept c = conceptService.getConceptByUuid("db1741f033144371a191649f98454148");
    Assert.assertEquals("Text", c.getDatatype());
    Assert.assertFalse(c.getRetired());
}
```
Anyway, what is this test about, and why is there an illegal UUID (`db1741f033144371a191649f98454148`)?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+		
+		loader.load();
+		{
+			// Verif by name
+			Context.setLocale(localeEn);
+			c = conceptService.getConceptByName("VACCINE MANUFACTURER");
```
mks-d, line 72: Change this to
```java
Concept c = conceptService.getConceptByName("VACCINE MANUFACTURER");
```
and assert each newly fetched concept in its own anonymous scope.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+			Assert.assertNotNull(c);
+			Assert.assertEquals(0, c.getDescriptions().size());
+			Assert.assertEquals("Finding", c.getConceptClass().getName());
+			Assert.assertEquals("Text", c.getDatatype().getName());
+			
+			// Verify by UUID
+			Context.setLocale(localeEn);
+			c = conceptService.getConceptByUuid("98b5c08e2069432da950878ccea81d19");
```
mks-d, line 80: Illegal UUID?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+			// Failed ones
+			Context.setLocale(localeEn);
+			Assert.assertNull(conceptService.getConceptByUuid("db2f5104-3171-11e7-93ae-92361f002670"));
+			Assert.assertNull(conceptService.getConceptByName("MEASLES VACCINATION"));
+			Assert.assertNull(conceptService.getConceptByName("db2f5460-3171-11e7-93ae-92361f002672"));
+			Assert.assertNull(conceptService.getConceptByName("HEPATITIS B VACCINATION"));
+			Assert.assertNull(conceptService.getConceptByUuid("00b29984-3183-11e7-93ae-92361f002679"));
+			
+			// Retired one
+			c = conceptService.getConceptByUuid("ed92472255674a4095559f2e0daad26d");
```
mks-d, line 114: Illegal UUID.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+			Assert.assertTrue(c.isRetired());
+			
+			// Un-retire one
+			Context.setLocale(localeEn);
+			c = conceptService.getConceptByUuid("f1f09c2f1e164804bd80f5df9e7f8dac");
```
mks-d, line 119: Illegal UUID.

mks-d (changes requested):
Thanks @suruchee, a couple of changes/comments.

After addressing them it would be great is to validate that everything works _with_ the Validator. Should we try to liaise about this, maybe over the TAC call? @gracepotma?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
ibacher: It would involve some refactoring to leverage something like that. Currently the module is calling `OpenConceptLabActivator.getDaemonToken()` where it needs a daemon token. In any case, the constructor for the activator for the OCL module doesn't actually do anything.

pom.xml
```diff
@@ -86,7 +86,7 @@
             <type>jar</type>
         </dependency>
 
-        <!-- JUnit, Powermock -->
+        <!-- JUnit, Powermock  -->
```
ibacher, line 89: This seems unnecessary?

api/src/main/java/org/openmrs/module/initializer/Domain.java
ibacher: @suruchee Per the discussion here, we should move this to before `CONCEPT` in the load order. Presumably this also means we need to renumber everything from `CONCEPT` -> `HTMLFORMS`.

suruchee:
@mks-d , I actually get confused with OCL UUID and Concept UUID, now I have updated with the required changes.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,130 @@
...
+			c = conceptService.getConceptByUuid("17AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
+			Assert.assertNotNull(c);
+			Assert.assertFalse(c.getRetired());
+			Assert.assertEquals("DIPTHERIA TETANUS BOOSTER", c.getFullySpecifiedName(localeEn).getName());
+			Assert.assertEquals("Drug", c.getConceptClass().getName());
+			Assert.assertEquals("N/A", c.getDatatype().getName());
+			
+		}
+		
+	}
```
mks-d, line 129: @suruchee this is what I mean with using anonymous scopes for each block of assertion, you can copy/paste the below snipptet:
```java
@Test
public void load_shouldImportOCLPackages() {
    // Verif setup
    {
        Concept c = conceptService.getConceptByUuid("1419AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
        Assert.assertEquals("Text", c.getDatatype());
        Assert.assertFalse(c.getRetired());
    }

    // Replay        
    loader.load();

    // Verif by name
    {
        Context.setLocale(localeEn);
        Concept c = conceptService.getConceptByName("VACCINE MANUFACTURER");
        Assert.assertNotNull(c);
        Assert.assertEquals(0, c.getDescriptions().size());
        Assert.assertEquals("Finding", c.getConceptClass().getName());
        Assert.assertEquals("Text", c.getDatatype().getName());
    }
    // Verify by UUID
    {
        Context.setLocale(localeEn);
        Concept c = conceptService.getConceptByUuid("3004BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB");
        Assert.assertNotNull(c);
        Assert.assertEquals("YELLOW FEVER VACCINATION", c.getName(localeEn).getName());
        Assert.assertEquals("Vaccine given for Yellow Fever.", c.getDescription().toString());
        Assert.assertEquals("Drug", c.getConceptClass().getName());
        Assert.assertEquals("N/A", c.getDatatype().getName());
    }
    // Verify in another locale
    {
        Context.setLocale(localeFr);
        Concept c = conceptService.getConceptByName("Fabricant du vaccin");
        Assert.assertNotNull(c);
        Assert.assertEquals(0, c.getDescriptions().size());
        Assert.assertEquals("Finding", c.getConceptClass().getName());
        Assert.assertEquals("Text", c.getDatatype().getName());;
    }
    // Verify just one name is enough
    {
        Context.setLocale(localeEn);
        Concept c = conceptService.getConceptByName("PNEUMOCOCCAL VACCINE");
        Assert.assertNotNull(c);
        Assert.assertEquals(3, c.getNames().size());
        Assert.assertEquals(0, c.getShortNames().size());
        Assert.assertEquals(0, c.getDescriptions().size());
        Assert.assertEquals("Drug", c.getConceptClass().getName());
        Assert.assertEquals("N/A", c.getDatatype().getName());;
    }
    // Failed ones
    {
        Context.setLocale(localeEn);
        Assert.assertNull(conceptService.getConceptByUuid("db2f5104-3171-11e7-93ae-92361f002670"));
        Assert.assertNull(conceptService.getConceptByName("MEASLES VACCINATION"));
        Assert.assertNull(conceptService.getConceptByName("db2f5460-3171-11e7-93ae-92361f002672"));
        Assert.assertNull(conceptService.getConceptByName("HEPATITIS B VACCINATION"));
        Assert.assertNull(conceptService.getConceptByUuid("00b29984-3183-11e7-93ae-92361f002679"));;
    }
    // Retired one
    {
        Concept c = conceptService.getConceptByUuid("83531AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
        Assert.assertTrue(c.isRetired());;
    }
    // Un-retire one
    {
        Context.setLocale(localeEn);
        Concept c = conceptService.getConceptByUuid("17AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
        Assert.assertNotNull(c);
        Assert.assertFalse(c.getRetired());
        Assert.assertEquals("DIPTHERIA TETANUS BOOSTER", c.getFullySpecifiedName(localeEn).getName());
        Assert.assertEquals("Drug", c.getConceptClass().getName());
        Assert.assertEquals("N/A", c.getDatatype().getName());
    }
    
}
```

mks-d (changes requested):
Thanks @suruchee.

I requested a change (I provided a snippet for it) on the main test class.
Also don't forget to address this: https://github.com/mekomsolutions/openmrs-module-initializer/pull/100#discussion_r615802730

I will have more comments on the tests though.

Finally we will have to look into proof testing all this with the validator as mentioned here: https://github.com/mekomsolutions/openmrs-module-initializer/pull/100#pullrequestreview-638589222

suruchee:
@mks-d, Thanks for the guidance, is the last update fine?

mks-d (dismissed):
@suruchee I'll have to get a closer look at the tests by actually looking at the JSON file loaded in the first place (which I haven't had a chance to do yet.)

In the meantime, please add a further commit with all README stuff as it is done for all other domains.

suruchee:
@Ruhanga, I have updated resolving the conflicts, will you please have a look?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+		}
+		
+		DaemonToken daemonToken = new DaemonToken("openconceptlab");
+		daemonTokens.put(daemonToken.getId(), daemonToken);
+		new OpenConceptLabActivator().setDaemonToken(daemonToken);
+	}
+	
+	@Test
+	public void load_shouldImportOCLPackages() {
+	    // Verif setup
```
Ruhanga, line 61: Let's use Setup instead.

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+	    {
+	        Concept c = conceptService.getConceptByUuid("1419AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
+	        Assert.assertEquals("Text", c.getDatatype());
+	        Assert.assertFalse(c.getRetired());
+	    }
+
+	    // Replay        
+	    loader.load();
+
+	    // Verif by name
```
Ruhanga, line 71: Typo ➙ Verify

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+	        Context.setLocale(localeEn);
+	        Concept c = conceptService.getConceptByName("PNEUMOCOCCAL VACCINE");
+	        Assert.assertNotNull(c);
+	        Assert.assertEquals(3, c.getNames().size());
+	        Assert.assertEquals(0, c.getShortNames().size());
+	        Assert.assertEquals(0, c.getDescriptions().size());
+	        Assert.assertEquals("Drug", c.getConceptClass().getName());
+	        Assert.assertEquals("N/A", c.getDatatype().getName());;
+	    }
+	    // Failed ones
```
Ruhanga, line 110: Verify failures

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+	    {
+	        Context.setLocale(localeEn);
+	        Assert.assertNull(conceptService.getConceptByUuid("db2f5104-3171-11e7-93ae-92361f002670"));
+	        Assert.assertNull(conceptService.getConceptByName("MEASLES VACCINATION"));
+	        Assert.assertNull(conceptService.getConceptByName("db2f5460-3171-11e7-93ae-92361f002672"));
+	        Assert.assertNull(conceptService.getConceptByName("HEPATITIS B VACCINATION"));
+	        Assert.assertNull(conceptService.getConceptByUuid("00b29984-3183-11e7-93ae-92361f002679"));;
+	    }
+	    // Retired one
```
Ruhanga, line 119: Verify retirement

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+	    {
+	        Concept c = conceptService.getConceptByUuid("83531AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
+	        Assert.assertTrue(c.isRetired());;
+	    }
+	    // Un-retire one
+	    {
```
Ruhanga, line 125: Verify un-retirement

Ruhanga (changes requested):

suruchee:
@Ruhanga, I have updated with your suggestions, will you please have a look?

Ruhanga (changes requested):
Hi @suruchee, thanks for the update. Could you also rebase on the PR?

Ruhanga (commented):
Could you also try the validator tool? Follow the documentation available [here](https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/validator.md).

suruchee:
@Ruhanga , I have already added the validator.

gracepotma:
@suruchee what's the next step here, or what do you need next from Nathan or others...?

Ruhanga:
@suruchee, looks like there there are still merge conflicts in the file `api-2.3/src/test/java/org/openmrs/module/initializer/api/loaders/LoadersOrderTest.java`. Could you have rebased and failed to push the changes?

suruchee:
Thanks @Ruhanga for the review, I have updated resolving the conflict, will you have a look again?

Ruhanga (commented):
@suruchee, does this branch build successfully locally your end? There seems to be some compilation error `openmrs-module-initializer/api/src/main/java/org/openmrs/module/initializer/api/loaders/OpenConceptLabLoader.java:[30,25] cannot access org.springframework.web.multipart.MultipartFile
`. Could you make sure the branch builds without errors?

suruchee:
@Ruhanga, this is a problem in ubuntu and only gets solved when spring framework dependency is added. But as you can also view from the previous conversation it is not recommended to use it. So, @ibacher, will you please have a look at it and help me solve it?

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+			throw new RuntimeException(e);
+		}
+
+=======
+		}
+		catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+		
+>>>>>>> 6cadc0cb071b7a85cda86a4f975797cdc47ec5e9
```
Ruhanga, line 60: @suruchee, could the problem be related to the conflicts not resolved here?

Ruhanga (changes requested):

api/src/test/java/org/openmrs/module/initializer/api/OpenConceptLabLoaderIntegrationTest.java
suruchee: Thank you @Ruhanga, I am sorry for missing out this file. I have updated after resolving the conflict. But the multipart issue still exists.

Ruhanga (commented):
@suruchee, could you point me to the commit/changes introducing the error?

suruchee:
@Ruhanga, it worked until https://github.com/mekomsolutions/openmrs-module-initializer/pull/100/commits/4b4551ee146c16ea01ccc86a1758d0321ef38b42 , then I removed spring dependency.

Ruhanga (commented):
@mks-d, @ibacher, I understand placing web related dependencies into the api layer as violating OpenMRS conventions. For some reason there is a hard requirement for this on [openconceptlab api](https://github.com/openmrs/openmrs-module-openconceptlab/blob/master/api/pom.xml#L24-L28). Maybe the fix should happen there or we have to the following dependency to iniz api;

```
      <dependency>
            <groupId>org.openmrs.web</groupId>
            <artifactId>openmrs-web</artifactId>
            <version>${openmrsPlatformVersion}</version>
            <scope>provided</scope>
      </dependency>
```

@suruchee, could you try with the above  added to `api/pom.xml` and ensure the tests pass?

mseaton:
> I understand placing web related dependencies into the api layer as violating OpenMRS conventions

Yes, I would recommend against this.  I haven't investigated, but it may be that we need to make changes to the ocl module in order to remove any web dependencies that it incorrectly has within it's api layer.

ibacher:
There are two things that the openconcept lab module gets from importing openmrs-web as a `provided` dependency:

1. Access to the `MultipartFile` class that's been mentioned here.
2. Access to Jackson to handle the JSON we get from OCL.

We already have several modules that depend on openmrs-web in the API part (as a `provided` dependency) in order to get access to the `MultipartFile` class, e.g., [attachments](https://github.com/openmrs/openmrs-module-attachments/blob/d2b421908419d6ce77696b5b81c9435a3f187fca/pom.xml#L65), [uiframework](https://github.com/openmrs/openmrs-module-uiframework/blob/d81c458c9bc5d55b4fbb1cecd60588f7f59d8b50/api/pom.xml#L28), and [htmlformentry](https://github.com/openmrs/openmrs-module-htmlformentry/blob/3839f633933ba96cabab6b538dd5b2049ae8ba5a/api/pom.xml#L36). I'm not meaning to suggest that this is the right way to go, just that this is how we've solved this need in the past.

For OpenMRS 1.9 (the earliest version we support in this module), Jackson is only available in openmrs-web as it wasn't moved to openmrs-api until 1.11 ([see TRUNK-3978](https://issues.openmrs.org/browse/TRUNK-3978)).

We could consider only supporting 1.11 or later instead of 1.9 or later, but that still doesn't help with the `MultipartFile` thing.

mseaton:
> We already have several modules that depend on openmrs-web in the API part

This was likely an inadvertent mistake that was made and we got away with it, but I don't think this is something we want to continue doing if we can avoid it.  Adding Jackson as a library to the initializer should be fine, right?

ibacher:
>  Adding Jackson as a library to the initializer should be fine, right?

It shouldn't be necessary to add Jackson to initializer or anything. Jackson is already available on the classpath at run time. I'm just explaining why it's a compile-time dependency for that module.

I suppose we can try to re-write things in the OCL module so we don't directly depend on `MultipartFile`, which would likely fix the issue here. I was just pointing out it's not a decision unique to the OCL module.

Ruhanga (commented):
Thanks @ibacher and @mseaton. I created a ticket to address this on the jira OCLOMRS project [here](https://issues.openmrs.org/browse/OCLOMRS-1052). Could you make it ready for for work? Thanks.

@suruchee would you then be in position to take it up?

suruchee:
Yes, sure @Ruhanga .

Ruhanga (commented):
@suruchee, are you able to progress with changes @ibacher has made on [OCLOMRS-1052](https://issues.openmrs.org/browse/OCLOMRS-1052)...

Ruhanga:
Hi @suruchee, how is this going?

suruchee:
Hi, @Ruhanga, I am working on it. I will update by monday.

pom.xml
```diff
@@ -61,7 +61,7 @@
         <idgenVersion>4.6.0</idgenVersion>
         <metadatasharingVersion>1.2.2</metadatasharingVersion>
         <metadatamappingVersion>1.3.4</metadatamappingVersion>
-
+        <openconceptlabVersion>1.2.6-SNAPSHOT</openconceptlabVersion>
```
ibacher, line 64: So, to take advantage of the new changes, this should be bumped to 1.2.8. I'll release that in a second.

suruchee:
@Ruhanga , I will update the code once the maven repo gets fixed https://status.io/pages/incident/534decd528328a247100004b/6148696c733c3e0537440673

Ruhanga (commented):
@suruchee, how is it going?

suruchee:
@Ruhanga, I have updated it, can you please review my last commit.

Ruhanga (commented):
@suruchee, did you notice errors in the tests such as [this](https://pastebin.com/ybL6UieZ)?

Ruhanga (commented):
Thanks for making progress on this @suruchee. How is it going at the moment with the tests?

suruchee:
@Ruhanga, I have tried adding few more tests and updating but the test still has some issues that I am not being able to solve. @ibacher, could you please help as I have already updated with the latest master?

ibacher:
@suruchee

Please verify that you're using the right data. For instance, the `load_shouldImportOCLPackages()` test tries to test for a concept with the UUID `3004BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB`. I'm not sure where that came from, but that's not a concept UUID (concept UUIDs in CIEL always end in `A`s; UUIDs ending in `B`s are concept names). The data in the Zip file can be found in [this gist](https://gist.github.com/ibacher/1cb11d8a2d63e983a03bdbd5686c5240).

I'm honestly not quite sure why the other tests are failing. Maybe it has something to do with the load order?

suruchee:
@ibacher, thanks for UUID idea and yes load order was causing issue, I have fixed load order and all tests but still I am getting a error which says:
Failed to load ApplicationContext getAttributeTypeByUuid_shouldGetAttributeType(org.openmrs.module.initializer.api.attributes.types.BahmniAttributeTypesProxyServiceTest): Failed to load ApplicationContext load_shouldLoadAccordingToCsvFiles(org.openmrs.module.initializer.api.BahmniAttributeTypesLoaderTest): Failed to load ApplicationContext

I don't think it's from OCL

suruchee:
@ibacher, thanks for UUID idea and yes load order was causing issue, I have fixed load order and all tests but still I am getting a error which says:
Failed to load ApplicationContext getAttributeTypeByUuid_shouldGetAttributeType(org.openmrs.module.initializer.api.attributes.types.BahmniAttributeTypesProxyServiceTest): Failed to load ApplicationContext load_shouldLoadAccordingToCsvFiles(org.openmrs.module.initializer.api.BahmniAttributeTypesLoaderTest): Failed to load ApplicationContext

I don't think it's from OCL

Ruhanga:
@suruchee, could you share the out put of a `mvn clean install`.

suruchee:
@Ruhanga, the output of mvn clean install is :
Tests in error: 
  saveAttributeType_shouldSaveAttributeType(org.openmrs.module.initializer.api.attributes.types.BahmniAttributeTypesProxyServiceTest): Failed to load ApplicationContext
  getAttributeTypeByUuid_shouldGetAttributeType(org.openmrs.module.initializer.api.attributes.types.BahmniAttributeTypesProxyServiceTest): Failed to load ApplicationContext
  load_shouldLoadAccordingToCsvFiles(org.openmrs.module.initializer.api.BahmniAttributeTypesLoaderTest): Failed to load ApplicationContext

ibacher:
@suruchee That's a different set of errors than I get with the code from this PR. Do you have local changes that aren't pushed yet? Also, it would be helpful to have more output than just the `Tests in error`, e.g., above them there should be tracebacks for where the errors actually occurred which is necessary to work out *why* those tests failed.

suruchee:
@ibacher, the output I am getting in mvn clean install is at https://pastebin.com/1beqLBLR, please have a look

Ruhanga (commented):
@suruchee do you get the same errors when you `mvn clean install` on the updated master branch?

suruchee:
@Ruhanga, the updated master branch is running fine. My test has some problems but it's hard to find.

mseaton:
> I am getting an error https://pastebin.com/1beqLBLR

@suruchee (cc @ibacher and @mks-d) - the key part of that error message is this:

`Caused by: java.lang.NoSuchMethodError: javax.persistence.Table.indexes()[Ljavax/persistence/Index;`

Googling this error, [you can find this link](https://stackoverflow.com/questions/50573767/hibernate-5-java-lang-nosuchmethoderror-javax-persistence-table-indexes) which is indicates the likely cause of the issue.

Indeed, running `mvn dependency:tree` on your branch reveals this conflicting library within the api-bahmni submodule:

```
[INFO] +- org.ict4h:atomfeed-client:jar:1.9.4:test
[INFO] |  \- org.ict4h:atomfeed-commons:jar:1.9.4:test
[INFO] |     \- javax.persistence:persistence-api:jar:1.0.2:test
```
To fix this problem, you should change api-bahmni/pom.xml and add an additional exclusion (after the existing rome exclusion) as follows:

        <dependency>
            <groupId>org.ict4h</groupId>
            <artifactId>atomfeed-client</artifactId>
            <version>${atomfeed-client.version}</version>
            <exclusions>
                <exclusion>
                    <groupId>rome</groupId>
                    <artifactId>rome</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>javax.persistence</groupId>
                    <artifactId>persistence-api</artifactId>
                </exclusion>
            </exclusions>
            <scope>test</scope>
        </dependency>

mseaton:
@suruchee , based on my building of your branch, once you resolve the above issue in the pom and in the LoadersOrderTest that I highlighted, you will also find that you get errors in some existing tests - notably the ConceptLoaderIntegrationTest and the ConceptSourcesLoaderIntegrationTest, presumably due to the fact that the OCL zip file you are importing is changing the expected data in these tests in a way that they are not expecting.  You'll need to work through these unless we find a way to better isolate these tests from each other (I may create a separate ticket for this - @mks-d FYI)

Ruhanga (commented):
@suruchee, how is it going?

suruchee:
Thanks @ibacher for the fix, all tests are passing now.

suruchee:
@Ruhanga, please have a look.

Ruhanga (commented):
Thanks very much @suruchee, @ibacher. The tests are passing now. @suruchee, did/could you test the `ocl` domain against the [initializer-validator](https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/validator.md)?

suruchee:
@Ruhanga yes it's working fine.

Ruhanga:
> test the ocl domain against the [initializer-validator](https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/validator.md)?


@suruchee, did you run the validator as documented in the link above?

Something along the lines below
```
mvn clean package -P validator
```
```
java -jar validator/target/initializer-validator-2.3.0-SNAPSHOT.jar --config-dir=./api/src/test/resources/testAppDataDir/configuration --domains='ocl'
```
Could you try running those command to test the validator against the `ocl` domain and share the results you get?

suruchee:
@Ruhanga, I got an error while running against ocl the output is https://pastebin.com/EhZWtuVZ, will you please suggest me the suggestion?

Ruhanga (commented):
@suruchee, a fix has been merged fixing the validator. Could you update this PR with the latest master changes and rerun the validator again? Please also provide the full logs from the run validator commands.

suruchee:
Hi @Ruhanga, thanks for the update. I have updated my branch with the latest master and ran the validator command but ended up with an error https://pastebin.com/nNBrPq9t please have a look

mseaton:
@mks-d / @Ruhanga / @suruchee - I think we need to prioritize merging this PR.  I've pulled this locally and run it, and I am able to build successfully, and I'm able to run the validator using the instructions that @Ruhanga linked to successfully (as far as I can tell, there i a lot of verbose output and warnings, but not obvious errors).

I also don't think getting every issue resolved with the validator is a blocker to getting this merged, and it is more important to get an initial pass at supporting OCL packages into a mainstream initializer release so that we can start actually testing it with real packages and real distributions.

If there are issues, we'll fix them, but they will be limited to the ocl domain anyway, so there is not much risk that we are introducing regressions here.

The code itself we are introducing in the Loader is dead simple - really just delegating to the OCL module, and supporting keeping that OCL zip in the configuration directories.

Can we merge this in ASAP please?

ibacher:
I'm in concurrence with @mseaton on this. The validator issues @suruchee indicated above aren't issues with validation per se, but issues with starting MariaDB4J itself (i.e., it's looking for the ncurses library which either isn't installed or the correct version isn't installed—realistically these binaries for MariaDB4J should be compiled completely statically if that's possible). Similarly on macOS, running the validator fails for me but because the command `mysql_install_db` fails with this message: "dyld: Library not loaded: /usr/local/opt/openssl/lib/libssl.1.0.0.dylib". Again a DLL not being found in the expected place.

Practically speaking, I've run a version of Iniz using these commits and used it to successfully load data from an OCL ZIP into an OpenMRS instance, which is the practical end of what we're trying to do so delaying this due to tooling issues seems... counter-productive.

ibacher:
Since this now as been updated to work with the validator, merging this in.

Issue #102, opened 2021-05-01
Support name and description on the programs domain


mseaton:
Currently the program domain does not support the name or description properties / columns on program, but rather always delegates the population of these to the associated Concept.

Implementations should be able to choose whether or not they prefer to populate these fields by adding explicit Name and/or Description columns.  These columns, if present, would supercede values from Concept.  These columns would be optional, and the current behavior would be the default if they are left unspecified.

mks-d:
Thanks a bunch @mseaton for your PR.

Issue #95, opened 2021-02-22
display:xy column to load UIFR-compatible i18n messages code and value pairs


mks-d:
#### 1) Background
REST WS is able to fill in a `display` property whose value is the L10n of the metadata. REST WS is trying several routes of L10n, and one of them is the UIFR pattern described [here](https://talk.openmrs.org/t/i18n-support-for-all-classes-extending-baseopenmrsmetadata/26218/25?u=mksd). This means that if a UIFR pattern-compliant message has been loaded then REST WS will populate `display` accordingly.
* Full reference: '[I18n support for all classes extending BaseOpenmrsMetadata](https://talk.openmrs.org/t/i18n-support-for-all-classes-extending-baseopenmrsmetadata/26218/24?u=mksd)'

#### 2) Acceptance Criteria
Let us take the 'locations' domain as an example.

| <sub>UUID</sub>                                 | <sub>Name</sub>           | <sub>Display:en</sub>        | <sub>Display:fr</sub>          | <sub>Display:kh</sub>          | <sub>Description</sub>                                                                                                                 | <sub>...</sub> |
|--------------------------------------|----------------|----------------|----------------|------------------|-----------------------------------------------------------------------------------------------------------------------------|-----|
| <sub>eb96c0d4-6248-476b-a079-0aaabfa2f614</sub> | <sub>Emergency Room</sub> | <sub>Emergency Room</sub> | <sub>Salle d'urgences</sub> | <sub>បន្ទប់​សង្គ្រោះ​បន្ទាន់</sub> | <sub>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</sub> | <sub>...</sub> |

This should ensure that the following i18n code-values are loaded in memory:
##### 1. For locale 'en':
```
ui.i18n.Location.name.eb96c0d4-6248-476b-a079-0aaabfa2f614=Emergency Room
```
##### 2. For locale 'fr':
```
ui.i18n.Location.name.eb96c0d4-6248-476b-a079-0aaabfa2f614=Salle d'urgences
```
##### 3. For locale 'kh':
```
ui.i18n.Location.name.eb96c0d4-6248-476b-a079-0aaabfa2f614=បន្ទប់​សង្គ្រោះ​បន្ទាន់
```

mks-d:
[...]

PR #103, opened 2021-05-01
[#102] Programs domain to support name and description headers.

mseaton:
I assume these were left off intentionally with the feeling that they should really be populated from Concept.  But we have existing program data that we are trying to migrate over to manage going forward with Iniz, and these have populated names and descriptions that vary from Concept names.

This also has the nice benefit of making the source CSV readable, as there are actually columns that describe what each row represents, rather than simply a collection of uuid references.

@mks-d let me know what you think.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/programs/ProgramLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/programs/ProgramLineProcessor.java
@@ -1,5 +1,6 @@
 package org.openmrs.module.initializer.api.programs;
 
+import org.apache.commons.lang.StringUtils;
 import org.openmrs.Concept;
 import org.openmrs.Program;
 import org.openmrs.api.ConceptService;
@@ -21,8 +22,7 @@ public class ProgramLineProcessor extends BaseLineProcessor<Program> {
 	private ConceptService conceptService;
 	
 	/**
-	 * @param headerLine The header line the processor will refer to.
-	 * @param service
+	 * @param conceptService
 	 */
 	@Autowired
 	public ProgramLineProcessor(@Qualifier("conceptService") ConceptService conceptService) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/initializer/api/programs/ProgramLineProcessor.java
```diff
@@ -35,9 +35,16 @@ public Program fill(Program program, CsvLine line) throws IllegalArgumentExcepti
...
+		if (StringUtils.isBlank(programName)) {
+			programName = Utils.getBestMatchName(programConcept, Context.getLocale());
+		}
 		program.setName(programName);
-		String proDescription = Utils.getBestMatchDescription(programConcept, Context.getLocale());
+		
+		String proDescription = line.getString(HEADER_DESC);
+		if (StringUtils.isBlank(proDescription)) {
+			proDescription = Utils.getBestMatchDescription(programConcept, Context.getLocale());
+		}
```
mks-d, line 47: I know that it's nothing to do with your PR, but could you use it to rename those guys `name` and `description` rather than `programName` and `proDescription`?

mks-d (commented):
Thanks @mseaton.

Definitely in support of this, especially because of https://github.com/mekomsolutions/openmrs-module-initializer/issues/95

api/src/main/java/org/openmrs/module/initializer/api/programs/ProgramLineProcessor.java
mseaton: Sure.  @mks-d I've changed those variable names.  Also updated the README.  Let me know how this looks.

readme/prog.md
```diff
@@ -8,13 +8,18 @@ programs/
...
+| <sub>Uuid</sub> | <sub>Void/Retire</sub> | <sub>Name</sub> | <sub>Description</sub> | <sub>Program concept</sub> | <sub>Outcomes concept</sub> | <sub>_order:1000</sub> |
 | - | - | - | - | - |
-| <sub>eae98b4c-e195-403b-b34a-82d94103b2c0</sub> | | <sub>TB Program</sub> | <sub>TB Program Outcomes</sub> | |
-
+| <sub>eae98b4c-e195-403b-b34a-82d94103b2c0</sub> | | TB Program | Program for tracking TB patients | <sub>Tuberculosis Treatment Program</sub> | <sub>TB Program Outcomes</sub> | |
 
 <br/>Let's review the headers.
 
+###### Header `Name`
+(Optional) If specified, this will be set to the name of the concept, otherwise the name will be inferred from the program concept name
```
mks-d, line 18: >(Optional) If specified, this will be set as the program name, otherwise the program name will be inferred from the program concept's name

readme/prog.md
```diff
@@ -8,13 +8,18 @@ programs/
...
+
+###### Header `Description`
+(Optional) If specified, this will be set to the description of the concept, otherwise the name will be inferred from the program concept description
```
mks-d, line 21: >(Optional) If specified, this will be set as the program description, otherwise the program description will be inferred from the program concept's description

mks-d (commented):
Minor readme changes.

Btw, shouldn't the same be done for workflows and states?

readme/prog.md
mseaton: Fixed

readme/prog.md
mseaton: Fixed

mseaton:
> Minor readme changes.
> 
> Btw, shouldn't the same be done for workflows and states?

One would think so, but if you look in the actual database (at least on my OpenMRS 2.3 instance) there are no name or description columns to be found.  If this has been changed in later versions of OMRS, we can add this in and presumably these fields will just continue to be ignored by the API as they have been.  I'm fine doing that in another follow-up commit / PR just to get this one in so we can start depending on it.

mks-d (approved):

PR #104, opened 2021-05-04
Leave existing non-specified attributes and location tags intact when loading

brandones:
Part of https://github.com/mekomsolutions/openmrs-module-initializer/issues/110

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
@@ -16,6 +16,8 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Component;
 
+import static org.apache.commons.lang3.BooleanUtils.isTrue;
+
 @Component("initializer.locationLineProcessor")
 public class LocationLineProcessor extends BaseLineProcessor<Location> {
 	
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
```diff
@@ -76,9 +76,9 @@ public Location fill(Location loc, CsvLine line) throws IllegalArgumentException
 			}
 		}
 		
-		loc.setTags(null);
 		String tags = line.getString(HEADER_TAGS, "");
 		if (!StringUtils.isEmpty(tags)) {
+			loc.setTags(null);
```
mseaton, line 81: I wouldn't think we would need this line moved here, since the line immediately after is also calling setTags.

api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
```diff
@@ -100,18 +100,21 @@ public Location fill(Location loc, CsvLine line) throws IllegalArgumentException
...
+				String tagName = StringUtils.removeStartIgnoreCase(header, HEADER_TAG_PREFIX);
+				LocationTag tag = locationService.getLocationTagByName(tagName);
+				if (tag == null) {
+					throw new IllegalArgumentException(
+					        "No location tag '" + tagName + "' exists for header '" + HEADER_TAG_PREFIX + tagName + "'");
+				}
+				Boolean value = line.getBool(header);
+				if (Boolean.TRUE.equals(value)) {
+					location.addTag(tag);
+				} else if (Boolean.TRUE.equals(location.hasTag(tagName))) {
```
mseaton, line 114: This feels a little awkward.  I'd probably just go with else { if (location.hasTag(tagName)) { ... } }

mseaton (commented):
Defer to @mks-d for approval overall.  A few comments from me just doing a cursory review.

api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
brandones:
```suggestion
```

api/src/main/java/org/openmrs/module/initializer/api/loc/LocationLineProcessor.java
brandones: Unfortunately `location.hasTag` returns a `Boolean` rather than a `boolean`, so it can be null.

mks-d: @brandones try this:
```java
import static org.apache.commons.lang.BooleanUtils.isTrue;

...

private void setLocationTagsFromPrefixHeaders(Location location, CsvLine line) {
    for (String header : line.getHeaderLine()) {
        if (StringUtils.startsWithIgnoreCase(header, HEADER_TAG_PREFIX)) {
            String tagName = StringUtils.removeStartIgnoreCase(header, HEADER_TAG_PREFIX);
            LocationTag tag = locationService.getLocationTagByName(tagName);
            if (tag == null) {
                throw new IllegalArgumentException("The location tag header '" + header
                        + "' references a location tag that does not exist: '" + tagName + "'.");
            }
            
            if (isTrue(line.getBool(header))) {
                location.addTag(tag);
            } else if (isTrue(location.hasTag(tagName))) {
                location.removeTag(tag);
            }
        }
    }
}
```
(I also edited the exception message.)

mks-d (changes requested):

brandones:
Incorporated your suggestion @mks-d

mks-d (commented):
Thanks @brandones I think I'm good with this. Just two things:
1. A line on the 2.2 release notes of the README.
2. Could you create an issue for this work to be expanded to all domains/headers? So that it's on the roadmap.

brandones:
Ok, documented!

mks-d (approved):

Issue #106, opened 2021-05-06
Initializer fails to install new programworkflowstate if it is retired


mseaton:
We have been migrating our existing metadata and configuration to Initializer, and we are currently migrating in our program metadata.  In this metadata, we have one program workflow state that is retired.  We wish this to be installed, but to be retired on the system.

If I upgrade an existing database in which this is already present, and Initializer is updating existing rows, things work fine.

However, if I start a new system from scratch, and Initializer needs to create a new Program Workflow State that is retired, it fails with a NullPointerException in the save method of the ProgramWorkflowStatesCsvParser, due to the fact that the associate ProgramWorkflow is null.

Looking into the code, this seems to be due to the fact that the logic in the CsvParser initializer method for each Line is to:

1. Bootstrap the object (load existing, or instantiate new)
2. Check if the Line is Retired/Voided, and if so, set to retired and move on without filling the rest of the object from the row
3. Otherwise, fill the object from the row 

The bootstrap logic in the ProgramWorkflowStatesCsvParser instance only loads an existing object from the database, or instantiates a new one and sets the UUID if supplied.  It does not set any of the required fields on the object - programWorkflow, concept, initial, terminal.  Because of this, if it is retired, these required fields are never set, and the parser attempts to save a ProgramWorkflowState with only the uuid and retired properties set.

I don't know if this is an issue across all other CsvParser implementations for other domains, and whether or not a holistic solution should be addressed, or whether this is limited to ProgramWorkflowStatesCsvParser.  @mks-d ?

mseaton:
Hmm.  So I ran into this exact same thing today while working on a different distribution, using the drugs domain, where we have a drugs.csv that has a bunch of retired drugs in it.  Setting this up into a new environment fails for all of the retired rows.  This seems like a systemic issue we need to fix @mks-d .  @brandones , @mogoodrich FYI

mseaton:
@Ruhanga / @mks-d .  Just a gentle nudge on this.  We remain blocked on using Iniz for several domains until we can get this resolved, and have a bunch of downstream work dependent on this which is sitting in PRs in other repos.  I could possibly fork the Mekom initializer repo and start a new release line to get us unblocked in the short term if that would be the most straightforward.  Thoughts?  @mogoodrich FYI.

PR #107, opened 2021-05-06
[#106] Parsers to fill new retired/voided entities.

mseaton:
It seems that the existing system is intentionally allowing CSV rows to constain just an identifier (uuid) and a void/retire flag set to true, and all other columns left blank, in order to allow voiding/retiring existing data without having full details.  Personally I don't love this feature, but it seems ingrained in the unit tests and there may be groups relying upon it, so I left it in place for backwards compatibility.

So, what this PR does is ensure that if the row represents a _new_ object being saved to the DB, that it _is_ populated with the full row details being saved, as if this does not happen then things fail badly on most objects.  The way I determined this was checking if an object has an id.  If an id exists, I am assuming it represents an existing loaded object.  There are a few domains for which this is not true - namely roles and privileges which do not have ids, and datafiltermappings which has it's own logic.  So I've implemented these in their subclasses.

I've added one unit test case that demonstrates the failure without the modifications, and with these modifications all of the unit tests pass (and I have tested this for my configurations that were running into this issue, and these all work).

@mks-d  / @Ruhanga / @rbuisson - I'd really appreciate a review and to try to get this merged in the next day if possible.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
@@ -236,7 +236,7 @@ public abstract class CsvParser<T extends BaseOpenmrsObject, P extends BaseLineP
 		//
 		// 2. Fill the instance using all line processors in their specified order, if indicated
 		//
-		if (shouldFillInstance(instance, csvLine)) {
+		if (shouldFill(instance, csvLine)) {
 			for (BaseLineProcessor<T> processor : lineProcessors) {
 				instance = processor.fill(instance, csvLine);
 				if (instance == null) {
@@ -256,11 +256,9 @@ public abstract class CsvParser<T extends BaseOpenmrsObject, P extends BaseLineP
 	}
 	
 	/**
-	 * @return true if a given instance should be filled with the contents of the csvLine By default, if
-	 *         an object is being voided or retired, and is not a new object, it is not filled from the
-	 *         row
+	 * @return true by default, or false for existing instances that are marked to be retired/voided.
 	 */
-	protected boolean shouldFillInstance(T instance, CsvLine csvLine) {
+	protected boolean shouldFill(T instance, CsvLine csvLine) {
 		boolean isVoidedOrRetired = BaseLineProcessor.getVoidOrRetire(csvLine);
 		return !(isVoidedOrRetired && instance.getId() != null);
 	}
--- a/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
@@ -1,12 +1,6 @@
 package org.openmrs.module.initializer.api;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.List;
-
+import com.opencsv.CSVReader;
 import org.apache.commons.lang.StringUtils;
 import org.openmrs.BaseOpenmrsData;
 import org.openmrs.BaseOpenmrsObject;
@@ -18,7 +12,12 @@ import org.openmrs.module.initializer.InitializerConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.opencsv.CSVReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 
 public abstract class CsvParser<T extends BaseOpenmrsObject, P extends BaseLineProcessor<T>> {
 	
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/openmrs/module/initializer/api/CsvParser.java
```diff
@@ -235,28 +234,37 @@ private T initialize(String[] line) throws APIException {
...
 		
 		return instance;
 	}
 	
+	/**
+	 * @return true if a given instance should be filled with the contents of the csvLine By default, if
+	 *         an object is being voided or retired, and is not a new object, it is not filled from the
+	 *         row
+	 */
+	protected boolean shouldFillInstance(T instance, CsvLine csvLine) {
```
Ruhanga, line 263: This made me go thinking whether to/not to address a use-case such as first having to update an old object then voiding it with the same `line` than having to edit it first and voiding it with two lines respectively, here. @mks-d, @rbuisson your thoughts...

Ruhanga: Ohh, sorry I missed this one. What I'm inquiring is whether it would be difficult for some reason to allow filling an old object and then retiring it?

mseaton: Well, it's just not the behavior that existed previously.  I'm trying to maintain as much of the existing behavior as possible in order to get this PR approved and to ensure backwards compatibility, while also fixing the bug we uncovered.  Previously, the method never filled the object at all if it was being retired.  Now, it fills it if it is new, but if it is not new then it continues not filling it at all.

We need to maintain this behavior (not filling the object if it is existing) because there are examples (in the unit tests at least, and I assume in reality) where the CSV row only contains the uuid of an object and a retired = true flag.  If we were to populate the existing object with this row, then we'd be nulling out the name, description, etc. and would get errors when trying to save the object due to missing required fields.

Ruhanga: Looks good to me.

mks-d, line 263: @mseaton could you rename the method to simply `shouldFill(T, CsvLine)`?

And change the JavaDoc to
>true by default, or false for existing instances that are marked to be retired/voided.

api/src/main/java/org/openmrs/module/initializer/api/privileges/PrivilegesCsvParser.java
```diff
@@ -54,6 +55,24 @@ public Privilege bootstrap(CsvLine line) throws IllegalArgumentException {
...
+	protected boolean shouldFillInstance(Privilege instance, CsvLine csvLine) {
+		boolean isVoidedOrRetired = BaseLineProcessor.getVoidOrRetire(csvLine);
+		if (!isVoidedOrRetired) {
+			return true;
+		}
+		Privilege existingPrivilege = userService.getPrivilegeByUuid(csvLine.getUuid());
+		if (existingPrivilege == null) {
+			String privilegeName = csvLine.get(PrivilegeLineProcessor.HEADER_PRIVILEGE_NAME, true);
+			existingPrivilege = userService.getPrivilege(privilegeName);
+		}
```
mks-d, line 72: I know it does it the other way around (name _then_ UUID), but could you use [`Utils#fetchPrivilege`](https://github.com/mekomsolutions/openmrs-module-initializer/blob/0cd4836c1d7ca167f647ddf87a26ed2c349128d6/api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java#L409-L415)?

api/src/main/java/org/openmrs/module/initializer/api/roles/RolesCsvParser.java
```diff
@@ -53,6 +54,23 @@ public Role bootstrap(CsvLine line) throws IllegalArgumentException {
...
+	@Override
+	protected boolean shouldFillInstance(Role instance, CsvLine csvLine) {
+		boolean isVoidedOrRetired = BaseLineProcessor.getVoidOrRetire(csvLine);
+		if (!isVoidedOrRetired) {
+			return true;
+		}
+		Role existingRole = userService.getRoleByUuid(csvLine.getUuid());
+		if (existingRole == null) {
+			existingRole = userService.getRole(csvLine.get(RoleLineProcessor.HEADER_ROLE_NAME, true));
+		}
```
mks-d, line 70: Same remark with [`Utils#fetchRole`](https://github.com/mekomsolutions/openmrs-module-initializer/blob/0cd4836c1d7ca167f647ddf87a26ed2c349128d6/api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java#L395-L401).

api/src/main/java/org/openmrs/module/initializer/api/privileges/PrivilegesCsvParser.java
mseaton: Discussed on slack with @mks-d and decided against this

mks-d (approved):

PR #112, opened 2021-05-18
[#115] Derive the name and order of each domain from the base enum.

mseaton:
#### What I have done:
Derive the name and order of each domain from the base enum, rather than maintain an independent name and order that needs to be manually maintained and updated.

@mks-d - I threw this together as a PR as it was easy enough to do as a means to ask "why don't we do this"?  If you want to keep it the way it is, that's fine.  I have the urge to do this every time I go to add a new domain and have to manually change all of the orders for all of the domains that fall after the one I'm trying to add.


mks-d (commented):
I like it! :-)

Mind you, this one will require a lot of README edits. But it simplifies something that was indeed unnecessarily complicated in the end.

mseaton:
Sounds good @mks-d .  How do you want me to proceed with this then?  Create a ticket to associate this with, and make the necessary readme changes?  Let me know.

mks-d (commented):
>Sounds good @mks-d . How do you want me to proceed with this then? Create a ticket to associate this with, and make the necessary readme changes? Let me know.

mks-d (commented):
>Sounds good @mks-d . How do you want me to proceed with this then? Create a ticket to associate this with, and make the necessary readme changes? Let me know.

@mseaton I would suggest to update the docs/READMEs as part of the PR Itself.

mseaton:
@mks-d that's fine, but I'm not 100% clear on what README changes you are referring to.  You seemed to imply that there would be a lot of changes required, but it's not obvious where.

mks-d (commented):
Sorry @mseaton you are right, this PR has no visible impact (which is obviously great), so there is no documentation to update. I somehow thought that some domains would see some noticeable changes, but not at all.

All good, except for the conflicts reported by GitHub.

mseaton:
@mks-d cool thanks.  I've merged master up and resolved the conflicts, so this should be good to go.

mks-d (commented):
Just the one-liner release note please, in the main README :-)

mseaton:
Done - feel free to edit to suit your tastes @mks-d

mks-d (approved):

Issue #118, opened 2021-05-27
Order Frequency domain will fail if any Orders with frequencies exist


mseaton:
Question here as to whether we should remove this validation from core, or whether we should fix Initializer to account for it.  But the core OrderServiceImpl class has this implementation for the saveOrderFrequency method:

```
	/**
	 * @see org.openmrs.api.OrderService#saveOrderFrequency(org.openmrs.OrderFrequency)
	 */
	@Override
	public OrderFrequency saveOrderFrequency(OrderFrequency orderFrequency) throws APIException {
		if (orderFrequency.getOrderFrequencyId() != null
				&& dao.isOrderFrequencyInUse(orderFrequency)) {		
			throw new CannotUpdateObjectInUseException("Order.frequency.cannot.edit");
		}
		
		return dao.saveOrderFrequency(orderFrequency);
	}
```

The Initializer OrderFrequenciesLoader / OrderFrequenciesCsvParser classes will save any OrderFrequency that it finds in the file, if the checksum of the entire file has changed.  This means that if we have an Order Frequency CSV, and then start entering orders that use these frequencies, if we ever add a new frequency to the file, this file will fail to load with the above error, even though we are not modifying any of the existing "in use" frequencies.

We had some special processing in our pre-Initializer codebase to handle this and are now realizing that Initializer does not account for this.  Thanks to @mogoodrich for noticing in code review.

@mks-d FYI

mseaton:
Personally I think this is lazy in the openmrs core code, and that the proper fix belongs in core where either this validation should be removed altogether, or should be improved to check that meaningful fields are actually being changed before throwing the error.  There are presumably legitimate cases where one would want to update an order frequency (to fix an incorrect configuration, or to point to a different (though equally or more correct) concept, etc.  Thoughts?

mseaton:
Here is my attempt at a unit test that demonstrates this issue.  If you run this, it will fail.  If you comment out the line that saves the encounter, it will pass.  https://github.com/mekomsolutions/openmrs-module-initializer/pull/119

mseaton:
I ended up changing OpenMRS core to allow for saving order frequencies that have been used.  See ticket here:
https://issues.openmrs.org/browse/TRUNK-6006

This is fixed in 2.3.4, 2.4.1, 2.5.0+.  @mks-d please feel free to close this issue if having this solved in those versions on-up is sufficient.

mks-d:
[...]

PR #119, opened 2021-05-27
Failing unit test that demonstrates that one cannot load an order fre…

mseaton:
…quency domain if any frequencies have existing orders associated with them.


mks-d:
Closed as per this: https://github.com/mekomsolutions/openmrs-module-initializer/issues/118#issuecomment-852046247

PR #125, opened 2021-07-12
Issue-95: Adding I18n support for openmrs metadata domains

Ruhanga:
Issue: https://github.com/mekomsolutions/openmrs-module-initializer/issues/95


Ruhanga (commented):
Hi @mks-d, you could review this draft PR when you get the chance.

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseCsvLoader.java
```diff
@@ -115,4 +119,50 @@ protected void load(InputStream is) throws Exception {
...
+						String[] codes = getLocalizationCodes(target.getClass().getSimpleName(), target.getUuid());
+						initializerMessageSource
+						        .addPresentation(new PresentationMessage(codes[0], displayLocale, display, null));
+						initializerMessageSource
+						        .addPresentation(new PresentationMessage(codes[1], displayLocale, display, null));
+					}
+				}
+			}
+		}
+	}
```
mks-d, line 148: It feels like this actual implementation of i18n-specific stuff should be done by `BaseI18nSupportLoader` no?

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseI18nSupportLoader.java
```diff
@@ -0,0 +1,54 @@
...
+ * A loader to focus on internationalization of {@link InputStream} entries through message
+ * properties loaded using {@link InitializerMessageSource}. This logic is executed after files of
+ * the domain have been loaded by {@link BaseFileLoader#loadUnsafe(List, boolean)} implementation.
+ * 
+ * @see {@link BaseInputStreamLoader}
+ * @see {@link BaseFileLoader}
+ * @see {@link Loader}
+ * @since 2.2.0
+ */
+public abstract class BaseI18nSupportLoader extends BaseInputStreamLoader {
```
mks-d, line 24: But even so, why not have a generic `DisplayHeaderLineProcessor` that would be added to the set of line processors for all CSV domains?

Why put this logic at the loader level?

Ruhanga: I envision there being some other non csv-file loaded domains which would have to extend `BaseI18nSupportLoader`.

api/src/test/java/org/openmrs/module/initializer/api/BaseI18nSupportLoaderIntegrationTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.openmrs.api.LocationService;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.initializer.Domain;
+import org.openmrs.module.initializer.DomainBaseModuleContextSensitiveTest;
+import org.openmrs.module.initializer.InitializerMessageSource;
+import org.openmrs.module.initializer.api.loaders.Loader;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+public class BaseI18nSupportLoaderIntegrationTest extends DomainBaseModuleContextSensitiveTest {
```
mks-d, line 20: Rather than having this ad-hoc test, could expand the tests on encounter types and locations then?

api/src/main/java/org/openmrs/module/initializer/api/DisplayLineProcessor.java
```diff
@@ -0,0 +1,65 @@
...
+		if (underscoreIndex > 0) {
+			shortClassName = shortClassName.substring(0, underscoreIndex);
+		} else {
+			underscoreIndex = shortClassName.indexOf("$");
+			if (underscoreIndex > 0) {
+				shortClassName = shortClassName.substring(0, underscoreIndex);
+			}
+		}
+		return shortClassName;
+	}
```
mks-d, line 64: Do you have some clues as to why this is needed?

If it is really needed, then move it as a `public static` in `Utils` and make it:
```java
public static String unProxy(String shortClassName) {
    shortClassName = StringUtils.substringBefore(shortClassName, "_$");
    shortClassName = StringUtils.substringBefore(shortClassName, "$");
    return shortClassName;
}
```
Then add a test case for it:
```java
@Test
public void unProxy_shouldReturnOriginalClassName() {
    Assert.assertEquals("EncounterType", unProxy("EncounterType$HibernateProxy$ODcBnusu"));
    Assert.assertEquals("EncounterType", unProxy("EncounterType_$$_javassist_26"));
}
```

mks-d (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/DisplayLineProcessor.java
Ruhanga: For the potential lazily-loaded hibernate objects, this is indeed needed to extrapolate the actual short class name.

api/src/test/java/org/openmrs/module/initializer/api/BaseI18nSupportLoaderIntegrationTest.java
Ruhanga: Indeed! Thanks, I missed this. I've added these changes.

api/src/test/java/org/openmrs/module/initializer/api/EncounterTypesLoaderIntegrationTest.java
```diff
@@ -102,5 +109,27 @@ public void load_shouldLoadOrderFrequenciesAccordingToCsvFiles() {
 			Assert.assertEquals("A new description for the oncology encounter.", et.getDescription());
 			Assert.assertEquals(us.getPrivilege("Can: View oncology encounter"), et.getViewPrivilege());
 		}
+		// verify EncounterTypes domain i18n
+		{
+			EncounterType et = es.getEncounterType("Triage Encounter");
+			String etUuid = et.getUuid();
```
mks-d, line 115: Just name this variable `uuid`, it's clear enough in the scope where it's defined.

api/src/test/java/org/openmrs/module/initializer/api/LocationsLoaderIntegrationTest.java
```diff
@@ -144,5 +150,28 @@ public void load_shouldLoadAccordingToCsvFiles() {
 			Location loc = ls.getLocationByUuid("2b9824a3-92f0-4966-8f34-1b105624b267");
 			Assert.assertNull(loc);
 		}
+		
+		// verify Locations domain i18n
+		{
+			Location loc = ls.getLocation("The Lake Clinic-Cambodia");
+			String locUuid = loc.getUuid();
```
mks-d, line 157: Same here.

api/src/test/resources/testAppDataDir/configuration/encountertypes/encountertypes.csv
```diff
@@ -1,7 +1,7 @@
-Uuid,Void/Retire,Name,Description,View privilege,Edit privilege,_order:1000
-,,Triage Encounter,An encounter for triaging patients.,,,
-aaa1a367-3047-4833-af27-b30e2dac9028,,Medical History Encounter,An interview about the patient medical history.,,,
-439559c2-a3a4-4a25-b4b2-1a0299e287ee,,X-ray Encounter,An encounter during wich X-rays are performed on the patient.,Can: View X-ray encounter,Can: Edit X-ray encounter,
-400d7e07-6de6-40ac-8611-dcce12408e71,,Foo Encounter,An foo encounter that should not be created.,Can: View foo encounter,,
-bed6f0f6-ab07-481f-929f-3d26e6cb1138,TRUE,,,,,
-,,Oncology Encounter,A new description for the oncology encounter.,Can: View oncology encounter,,
\ No newline at end of file
+Uuid,Void/Retire,Name,Description,View privilege,Edit privilege,display:en,display:km_KH,_order:1000
+,,Triage Encounter,An encounter for triaging patients.,,,Triage Encounter (translated),ទ្រីយ៉ាហ្គេនស៊ើរ,
```
mks-d, line 2: What happens btw when a display value is left empty? Do we have any coverage for this?

mks-d (commented):
README? :-)

Otherwise it's looking great!

api/src/test/resources/testAppDataDir/configuration/encountertypes/encountertypes.csv
Ruhanga: The message property is not loaded. I've added a test to verify this.

Ruhanga (commented):
@mks-d, it's now that I remember one of the reasons for the previous approach. Perhaps we could have a work around with this one too. That the messages have to be loaded always irrespective of the checksums being computed. The approach we have now taken is smart but I'd think loads the messages once given there is no valid checksum for the csv file being loaded. Would you have any suggestions or options for a workaround on this?

api/src/main/java/org/openmrs/module/initializer/api/DisplayLineProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+		LocalizedHeader l10nHeader = LocalizedHeader.getLocalizedHeader(line.getHeaderLine(), HEADER_DISPLAY);
+		
+		l10nHeader.getLocales().stream().forEach(locale -> {
+			String display = line.get(l10nHeader.getI18nHeader(locale));
+			if (!StringUtils.isEmpty(display)) {
+				String shortClassName = Utils.unProxy(instance.getClass().getSimpleName());
+				msgSource.addPresentation(new PresentationMessage(
+				        "ui.i18n." + shortClassName + ".name." + instance.getUuid(), locale, display, null));
+				msgSource.addPresentation(new PresentationMessage("org.openmrs." + shortClassName + "." + instance.getUuid(),
+				        locale, display, null));
```
mks-d, line 43: Just to be clear is `"org.openmrs." + shortClassName` really what is supposed to be done?

So for instance an entity `org.openmrs.foo.Bar`:
* would lead to the short name `Bar`
* that would in turn lead to a message key prepended with `org.openmrs.Bar` (without the `.foo`)

Correct?

mseaton: That does not sound like the behavior we want to me.

mks-d (commented):
>it's now that I remember one of the reasons for the previous approach. Perhaps we could have a work around with this one too. That the messages have to be loaded always irrespective of the checksums being computed. The approach we have now taken is smart but I'd think loads the messages once given there is no valid checksum for the csv file being loaded. Would you have any suggestions or options for a workaround on this?

Oh I see, very good point... I think that if this has to be linked to the checksum lifecycle then we may have to persist the localisation strings and introduce a special loader that always reads from this persisted localisation data:

1. The usual loaders would persist the localisation data (instead of adding localisation data as presentations to the message source as it is done now in the current PR). This persisted localisation data would be updated on a checksum basis, as per usual.
2. A new special i18n loader would then read what is persisted and add the in-memory presentations to the message source. And that special loader would presumably always load last and bypass the checksums mechanism.

What needs to be persisted is simple on the principle: a UUID and a map locale-to-L10n string.
Because it is maybe a pain to persist maps, we could simply flatten them as a `JSONObject` or whatever marshalling we prefer.

@ibacher @mseaton what do you think?

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -552,4 +552,14 @@ public static boolean isAppointmentLocation(Location location) {
...
 	}
+	
+	/*
+	 * Turns a proxy short class name into the original short class name.
+	 * Eg. "EncounterType$HibernateProxy$ODcBnusu" or "EncounterType_$$_javassist_26" → "EncounterType"
+	 */
+	public static String unProxy(String shortClassName) {
+		shortClassName = StringUtils.substringBefore(shortClassName, "_$");
+		shortClassName = StringUtils.substringBefore(shortClassName, "$");
+		return shortClassName;
```
mseaton, line 563: There is a method in OpenMRS core:
HibernateUtil.getRealObjectFromProxy(object)
I'd recommend that we use that here.

api/src/main/java/org/openmrs/module/initializer/api/DisplayLineProcessor.java
mseaton: @mks-d / @Ruhanga - see my other comment about HibernateUtil's possible use here.

mseaton:
@mks-d and @Ruhanga - that seems reasonable.  Would seem to make sense to persist the generated message properties in the same format as is expected by initializer, maybe in a reserved location or namespace, or in memory...

ibacher:
@mks-d That sounds like a good idea to me. One question: would it make sense to just construct the properties entry and dump that to a properties file directly? I think that's the most straight-forward way to load something into a `MessageSource` with very little code on our end.

api/src/main/java/org/openmrs/module/initializer/api/DisplayLineProcessor.java
mks-d: @mseaton I was aware of `HibernateUtil`, however I thought that this was good enough and more lightweight than a full unproxying. But I leave you as the final judge on this point.

mks-d (commented):
@ibacher this is not how Iniz has been loading messages until now, see here: https://github.com/mekomsolutions/openmrs-module-initializer/blob/fa9f236e43398837ca3cdc5f47b2fde1af3318b4/api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java#L121-L151
Admittedly this is old code, and there may be room for improvements ;-) But just to say that it adds messages one by one to a cache, that is then processed by the message source.

But still the main point is rather about solving the issue of the checksums lifecycle, we need all CSVs to be processed as usual. As usual means that the outcome is persisted. And then this persisted outcome should be systematically dumped into the message source.

This will be complicated to do with one properties file, it looks like we need to persist the localisation data entity by entity.

ibacher:
@mks-d Well, if you've already got that around, might as well re-use it 😁 . I think the proposal otherwise sounds fine.

We probably want to have some mechanism that can link from the persisted message store back to the source CSV that created it (e.g., a naming convention for the persisted file or some kind of metadata).

PR #126, opened 2021-07-20
Loading should fail when a concept is specified but not found: program outcomes

brandones:
Loading should fail when a concept is specified but not found. Previously, if a program outcome was provided which did not correspond to any concept, the program would be saved with outcome == null and no error would be registered.

This behavior should be made consistent as we find other places where OpenMRS allows null concepts.

This also moves the `findConceptByName` call after the by-mapping call, since former always logs a warning when a concept is not found. This means that lots of warnings might be logged for concepts that are specified by mapping.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/programs/ProgramLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/programs/ProgramLineProcessor.java
@@ -47,8 +47,11 @@ public class ProgramLineProcessor extends BaseLineProcessor<Program> {
 		}
 		program.setDescription(description);
 		
-		Concept outcomeConcept = Utils.fetchConcept(line.get(HEADER_OUTCOMES_CONCEPT), conceptService);
-		program.setOutcomesConcept(outcomeConcept);
+		String outcomeConceptString = line.get(HEADER_OUTCOMES_CONCEPT);
+		if (StringUtils.isNotBlank(outcomeConceptString)) {
+			Concept outcomeConcept = Utils.fetchConcept(outcomeConceptString, conceptService, true);
+			program.setOutcomesConcept(outcomeConcept);
+		}
 		
 		return program;
 	}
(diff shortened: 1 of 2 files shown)


mogoodrich (approved):
LGTM@

brandones:
@mks-d

brandones:
Ping @mks-d

api/src/main/java/org/openmrs/module/initializer/api/programs/ProgramLineProcessor.java
```diff
@@ -47,8 +47,11 @@ public Program fill(Program program, CsvLine line) throws IllegalArgumentExcepti
 		}
 		program.setDescription(description);
 		
-		Concept outcomeConcept = Utils.fetchConcept(line.get(HEADER_OUTCOMES_CONCEPT), conceptService);
-		program.setOutcomesConcept(outcomeConcept);
+		String outcomeConceptString = line.get(HEADER_OUTCOMES_CONCEPT);
+		if (StringUtils.isNotBlank(outcomeConceptString)) {
+			Concept outcomeConcept = Utils.fetchConcept(outcomeConceptString, conceptService, true);
+			program.setOutcomesConcept(outcomeConcept);
+		}
```
mks-d, line 54: Could this be sufficient:
```java
Concept outcomeConcept = Utils.fetchConcept(line.get(HEADER_OUTCOMES_CONCEPT), conceptService);
program.setOutcomesConcept(Optional.of(outcomeConcept).orElseThrow(() -> new IllegalArgumentException("...")));
```
?
In other words, could this be handled on a case by case basis in each line processor?

brandones: Sure, it could. Is there an example of a kind of concept for which we would not want to fail hard if the user specifies a non-existent concept?

brandones: Oh that's right, it's only important for nullable concepts. I don't know of other examples, off the top of my head. Perhaps drug concept? Anyway I think you make a fair point.

brandones:
Stale; closing.

PR #128, opened 2021-07-21
fetchProgram* functions should ignore retired. Fixes #127

### LINKED ISSUE #127: Program workflow state loader fails to find unique non-retired workflow
The workflow state loader looks up a workflow by concept, using the function [fetchProgramWorkflow](https://github.com/mekomsolutions/openmrs-module-initializer/blob/752edb1b5371208d6db0020de65c8c02db6855de/api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java#L354).

This function calls [getProgramWorkflowsByConcept](https://github.com/openmrs/openmrs-core/blob/f76ee93cc46271997e577ed7fa223ac4a4e53dcb/api/src/main/java/org/openmrs/api/db/hibernate/HibernateProgramWorkflowDAO.java#L380), which returns all workflows matching that concept, both retired and un-retired.

`fetchProgramWorkflow` returns `null` if more than one workflow is found. Problematically, it will return `null` when two workflows are found, which can happen if one of them is retired.

brandones:


brandones:
@mks-d @mogoodrich

mseaton (changes requested):
See my comments on the ticket, but I don't believe this is the behavior we want.  Retired <> Voided.

PR #129, opened 2021-08-18
[#65] Support for RelationshipType.

brandones:
Fixes https://github.com/mekomsolutions/openmrs-module-initializer/issues/65

Building on the work started by @samuelmale .

@mks-d @Ruhanga


brandones:
Note that I'm not sure we should even include support for "weight" and "preferred." I'm waiting to hear back on https://talk.openmrs.org/t/preferred-relationship-types/34418 .

api/src/test/java/org/openmrs/module/initializer/api/RelationshipTypesLoaderIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+		// a relationship type for edit
+		{
+			RelationshipType rt = new RelationshipType();
+			rt.setName("Supervisor/Supervisee");
+			rt.setDescription("An old description for supervisor to supervisee relationship");
+			rt.setUuid(RELATIONSHIP_TYPE_TO_EDIT_UUID);
+			rt.setaIsToB("Supervisor");
+			rt.setbIsToA("Supervisee");
+			ps.saveRelationshipType(rt);
+		}
```
mks-d, line 47: We tend to prefer test metadata created through the appropriate XML file in [here](https://github.com/mekomsolutions/openmrs-module-initializer/tree/master/api/src/test/resources/testdata), most likely simply by adding to **test-metadata.xml**.

README.md
```diff
@@ -165,6 +167,7 @@ https://github.com/mekomsolutions/openmrs-module-initializer/issues
 * Bulk creation and edition of Bahmni appointment service types provided through CSV files in **configuration/appointmentservicetypes**.
 * Renaming domain: `appointmentsservicesdefinitions` → `appointmentservicedefinitions`.
 * Renaming domain: `appointmentsspecialities` → `appointmentspecialities`.
+* Bulk creation and edition of relationship types provided through CSV files in **configuration/relationshiptypes*
```
mks-d, line 170:
```
Bulk creation and edition of relationship types provided through CSV files in **configuration/relationshiptypes**.
```
(You missed the last two chars: "`*.`".)

mks-d (commented):
Thanks so much @brandones. This looks very good, you are definitely an agile Iniz dev by now ;-)
I just left some minor comments.

api/src/test/resources/testdata/test-metadata.xml
```diff
@@ -48,7 +48,10 @@
   <idgen_seq_id_gen id="2" first_identifier_base="1" next_sequence_value="2" base_character_set="0123456789" prefix="" min_length="0" max_length="0" />
   
   <idgen_auto_generation_option id="15000" source="1" identifier_type="1" manual_entry_enabled="false" automatic_generation_enabled="true" uuid="eade77b6-3365-47ed-9ee3-2324598629eb" />
-    
+
+  <relationship_type relationship_type_id="1" a_is_to_b="Aunt" b_is_to_a="Niece" preferred="false" weight="0" description="A relationship of an aunt and her niece" creator="1" date_created="2007-05-04 00:00:00.0" retired="false" uuid="3982f469-cedc-4b2d-91ea-fe38f881e1a0"/>
+  <relationship_type relationship_type_id="2" a_is_to_b="Supervisor" b_is_to_a="Supervisee" preferred="false" weight="0" description="An old description for supervisor to supervisee relationship" creator="1" date_created="2007-05-04 00:00:00.0" retired="false" uuid="53d8a8f3-0084-4a52-8666-c655f5bd2689"/>
```
mks-d, line 53: Just to avoid interferences with test relationship types from other datasets (such as Core's standard dataset), maybe set "unpredictable" IDs here such as:
```xml
<relationship_type relationship_type_id="15001" ... >
<relationship_type relationship_type_id="15002" ... >
```

api/src/main/java/org/openmrs/module/initializer/api/relationship/types/RelationshipTypeLineProcessor.java
```diff
@@ -0,0 +1,37 @@
...
+		instance.setbIsToA(line.get(B_IS_TO_A, true));
+		
+		Boolean preferred = line.getBool(PREFERRED);
+		if (preferred != null) {
+			instance.setPreferred(line.getBool(PREFERRED));
+		}
+		Integer weight = line.getInt(WEIGHT);
+		if (weight != null) {
+			instance.setWeight(weight);
+		}
```
mks-d, line 33: So to be clear here, in fact `preferred` and `weight` are mandatory in the sense that [they are not nullable](https://github.com/openmrs/openmrs-core/blob/3f06bed8a6fca90248d050275dcdb96b86b1350f/api/src/main/resources/org/openmrs/api/db/hibernate/RelationshipType.hbm.xml#L36-L38), but via Iniz you de facto make them optional.

The question is, if someone edits the `preferred` member of a rel. type through an OpenMRS config. by leaving it blank in the CSV, what is the intent? "Clear the value"? That's not possible since its not nullable, then what does it mean, set it back to its default value? Because I guess the intent is not to error. If so, how to get back to that default, like this → `(new RelationshipType()).getPreferred()`?
```java
Boolean preferred = line.getBool(PREFERRED);
instance.setPreferred(preferred != null ?
  preferred : (new RelationshipType()).getPreferred());
```

I'm certainly overthinking this specific case but it would be interesting to see what others (@bmamlin, @djazayeri, @ibacher, @mseaton) think as I suspect that this kind of stuff may have not been implemented properly in other places...

dkayiwa: > by leaving it blank in the CSV, what is the intent? "Clear the value"?

I do not think one would intentionally clear a non nullable field. So it sounds safe to assume that it was done accidentally. And in that case, restoring it to the default (`RelationshipType. getPreferred(),  RelationshipType.getWeight()`) makes sense.

brandones: Sure, that seems reasonable to me. Though if the CSV doesn't have a header for "weight" or "preferred," I'd argue that those fields should just be left alone for entities being edited.

ibacher: @brandones I think that's definitely the right behaviour when the columns are *omitted*. How would you expect things to work if there was a "weight" column but its left blank for one or two entries rather than omitted entirely?

ibacher: In the later case, I can see arguments for either leaving the alone or resetting them to default. Resetting them to the default value strikes me as somewhat more in line with Iniz's WYSIWYG idea (because then the same row means the same thing whether the row is a new piece of metadata or an existing piece of metadata).

mks-d: Thanks @dkayiwa.

@brandones yes sure, I was only pointing to the case of explicitly setting a blank value through the CSV (not the case of the missing column altogether). I don't think that `getBool` lets you distinguish between those two cases, which probably means that we will have to let it for now.

mks-d (approved):

PR #131, opened 2021-08-25
Remove 'name' attribute from relationship type

brandones:
@mogoodrich pointed out that the "name" attribute of relationship types isn't actually stored in the database.

Attn @mks-d

### DIFF
--- a/api/src/test/java/org/openmrs/module/initializer/api/RelationshipTypesLoaderIntegrationTest.java
+++ b/api/src/test/java/org/openmrs/module/initializer/api/RelationshipTypesLoaderIntegrationTest.java
@@ -29,7 +29,6 @@ public class RelationshipTypesLoaderIntegrationTest extends DomainBaseModuleCont
 		{
 			RelationshipType rt = ps.getRelationshipTypeByUuid("c86d9979-b8ac-4d8c-85cf-cc04e7f16315");
 			Assert.assertNotNull(rt);
-			Assert.assertEquals("Uncle/Nephew", rt.getName());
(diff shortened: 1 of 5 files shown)


api/src/test/java/org/openmrs/module/initializer/api/RelationshipTypesLoaderIntegrationTest.java
```diff
@@ -29,7 +29,6 @@ public void load_shouldLoadRelationshipTypesAccordingToCsvFiles() {
 		{
 			RelationshipType rt = ps.getRelationshipTypeByUuid("c86d9979-b8ac-4d8c-85cf-cc04e7f16315");
 			Assert.assertNotNull(rt);
-			Assert.assertEquals("Uncle/Nephew", rt.getName());
```
mks-d, line 32: >the "name" attribute of relationship types isn't actually stored in the database

@mogoodrich / @brandones how come such asserts (that you have suggested to remove) would have worked until now then?

If you look at the test above, it's fetching a relationship type by UUID and that fetched relationship type does carry a name...

brandones: I would guess that that's an artifact of the test environment. Or something related to the Hibernate cache. It is easy enough to verify that the database model has no `name` attribute: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/resources/org/openmrs/api/db/hibernate/RelationshipType.hbm.xml

brandones:
Stale; closing.

Issue #130, opened 2021-08-20
Add providerrole domain for ProviderManagement module support


brandones:
We should add a new providerroles domain to Initializer. This will require an aware-of dependency to Provider Management module.

https://github.com/openmrs/openmrs-module-providermanagement

brandones:
Heads up @mks-d , if you don't have any objections I'm going to implement this shortly.

mks-d:
This guy from **openmrs-module-providermanagement** right: [`ProviderRole`](https://github.com/openmrs/openmrs-module-providermanagement/blob/19d69d703ebb20f4954fd69fddf4b8f05623ad73/api/src/main/java/org/openmrs/module/providermanagement/ProviderRole.java#L35)?

If so, no objections. Thanks for the heads up @brandones.

mseaton:
@mks-d and @brandones - I have a PR ready to add this feature.  Thanks for any time you can spend reviewing and helping to get this merged in.

mseaton:
This has been committed per above.

PR #134, opened 2021-09-02
[#130]: Add Provider Roles domain

mseaton:


gracepotma:
Nice! Thank you @mseaton

Issue #135, opened 2021-09-08
Add support for "version" to concept domain


mseaton:
We have a handful of concepts where we have populated the "version" property on Concept.  This property is settable via the 1.x concept management interface, and is a non-deprecated property of Concept.  Initializer should support setting this property optionally.

mseaton:
@mks-d - I have merged this PR following @Ruhanga's approval and our Slack discussion.  Please let me know if any post-commit updates are necessary prior to release.

PR #147, opened 2021-10-19
[#135] Add support for "version" to concepts domain.

mseaton:
Straight-forward addition to the "concepts" domain that allows setting the "version" property on Concept.


Ruhanga (approved):
Thanks @mseaton, LGTM.

Issue #155, opened 2021-11-05
Message properties do not always return correct messages


mseaton:
We have recently been using the messageproperties domain quite a bit as a means to deliver all of our translations across our distributions.  Over the course of this time, we have noticed some inconsistent behavior, and finally had a chance to take a look at it.  There are a few issues that we have discovered.  In summary those issues are:

1. The InitializerMessageSource does not default to the system locale if no message is found in the specified locale.  Specifically, if there is a translation present in messages_en.properties, and this translation is not present in messages_fr.properties, when someone using the system in French views this message they will see the message code rather than the English translation.  This is inconsistent with the core Message Source (and what we typically expect).

2. The InitializerMessageSource does not correctly handle a situation where the parent source has a "better" translation than the InitializerMessageSource.  For example, if the parent (the OpenMRS message source) has a message code defined for "fr_FR", and the Initializer source has the same message code defined for "fr", when someone uses the system in the "fr_FR" locale, they will get the less-specific Initializer "fr", rather than the more-specific parent translation.

The InitializerMessageSource needs to be modified to account for these issues.  Having spent some time looking at the code in depth, my belief is that the best course of action for a solution is to write a comprehensive Message Source that does not defer to the OpenMRS message source as parent, but rather re-implements the OpenMRS message source with additional Initializer functionality included.  If successful, this could ultimately be harvested back into OpenMRS core in a way that would allow Initializer (and other modules) to plug into it and extend it.

@mks-d FYI.

Ruhanga:
Thanks @mseaton for looking into this. No objections from me.

mogoodrich:
I haven't reviewed, but your proposal makes sense to me.  To clarify, when you say "system locale" do you mean the locale defined by the "default_locale" global property?

mseaton:
> To clarify, when you say "system locale" do you mean the locale defined by the "default_locale" global property?

No, I mean whatever is returned by Locale.getDefault() from the JDK, which I assume is a server locale.  This is what I've seen used elsewhere.

But this does bring up a good point, and I had also been thinking something along these terms in relation to the overall message cache that we maintain in memory - namely, should this use the default_locale global property as the "fallback" locale instead of (or prior to) using Locale.getDefault()?  And should we only actually worry about processing message codes that match those in the locale.allowed_list global property?  (No use keeping messages for dozens of languages the server will never use in memory).  Interested in your thoughts @mogoodrich and @mks-d ...

PR #156, opened 2021-11-05
[#155] Overhaul of InitializerMessageSource.

mseaton:
This PR undertakes a reasonably sizeable overhaul of the InitializerMessageSource.  The initial impetus for this was to investigate and fix issues related to appropriate "fallback" messages not being rendered when expected.  For example, we had situations where message codes were being rendered even when messages exist for the "default locale" (i.e. English).  Upon starting to investigate, we encountered a number of other examples of unexpected behavior, which we attributed to limitations in the existing OpenMRS and Spring message source implementations around precedence of message loading in both Spring and Openmrs message sources, and our delegation of message loading between custom source and parent source.

Ultimately, we concluded that the most effective means of ensuring consistent, expected behavior was to no longer split message resolution responsibility between custom source and parent source, but to fully implement a new message source to cover all known use cases.  Thus, this new message source is a replacement for the OpenMRS message source, not a child to it.  It takes reponsibility for loading messages first from the OpenMRS core classpath, then from each started module classpath, and finally from Initializer domains.

This new message source also adds in improved support for fallback locales, using the OpenMRS LocaleUtility.getLocalesInOrder() function to fallback to configured default and supported locales, before finally falling back to the system locale.  The new message source also allows for other language-specific fallbacks to be configured as needed - i.e. one can now configure that the "ht" Locale should use "fr" as a fallback language before falling back as described above.

@mks-d, @Ruhanga, @mogoodrich, @rbuisson - please let me know your thoughts and how we can move this forward.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
+++ b/api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
@@ -51,19 +51,19 @@ import java.util.concurrent.ConcurrentHashMap;
 /**
  * Custom Message Source that is intended to replace the OpenMRS core message source and provide
  * enhanced capabilities. This message source loads in messages from OpenMRS core, followed by those
- * defined in each module in the order in which each is started, and finally by those defined in configured
- * Initializer domains. Messages are loaded in this order to enable predictable and expected
- * overrides. Thus, messages defined with the same code and Locale in Initializer domains will take
- * first precedence, followed by those in modules, and finally by those in OpenMRS core. Ultimately,
- * a Map backs this source, with Locale and message code as keys to the Map, and thus values loaded
- * later will replace those loaded earlier in the process. The requested Locale always takes
- * precedence over the order of sources. For example, if a message is requested for Locale 'fr_FR',
- * all sources are first searched for a match to 'fr_FR' prior to then searching for 'fr' across all
- * sources. If no match is found in a given Locale, then the fallback is to search for the best
- * match in the System Locale. If no match is found in the System Locale, then the message code
- * itself is returned as a final fallback. This source allows for supporting additional fallback
- * languages, as well as defining additional classpath patterns and domains to search for message
- * property files. See method javadoc for further details.
+ * defined in each module in the order in which each is started, and finally by those defined in
+ * configured Initializer domains. Messages are loaded in this order to enable predictable and
+ * expected overrides. Thus, messages defined with the same code and Locale in Initializer domains
+ * will take first precedence, followed by those in modules, and finally by those in OpenMRS core.
+ * Ultimately, a Map backs this source, with Locale and message code as keys to the Map, and thus
+ * values loaded later will replace those loaded earlier in the process. The requested Locale always
+ * takes precedence over the order of sources. For example, if a message is requested for Locale
+ * 'fr_FR', all sources are first searched for a match to 'fr_FR' prior to then searching for 'fr'
+ * across all sources. If no match is found in a given Locale, then the fallback is to search for
+ * the best match in the System Locale. If no match is found in the System Locale, then the message
+ * code itself is returned as a final fallback. This source allows for supporting additional
+ * fallback languages, as well as defining additional classpath patterns and domains to search for
+ * message property files. See method javadoc for further details.
  * 
  * @see <a href=
  *      "https://talk.openmrs.org/t/address-hierarchy-support-for-i18n/10415/19?u=mksd">...</a>
@@ -182,8 +182,7 @@ public class InitializerMessageSource extends AbstractMessageSource implements M
 			Properties properties = resources.get(resourceName);
 			String nameWithoutExtension = FilenameUtils.removeExtension(resourceName);
 			Locale locale = getLocaleFromFileBaseName(nameWithoutExtension);
-			String baseName = StringUtils.removeEnd(nameWithoutExtension, "_" + locale.toString());
-			log.trace("Adding " + properties.size() + " messages from " + baseName + " in locale: " + locale);
+			log.trace("Adding " + properties.size() + " messages from " + resourceName + " in locale: " + locale);
 			for (Object property : properties.keySet()) {
 				String key = property.toString();
 				String value = properties.getProperty(key);
@@ -341,18 +340,20 @@ public class InitializerMessageSource extends AbstractMessageSource implements M
 	public List<String> getClasspathPatternsToScan() {
 		return classpathPatternsToScan;
 	}
-
+	
 	/**
-	 * One can use this method to add additional classpath resource patterns to scan, if necessary
-	 * Note, that one will need to ensure that the refreshCache() method is invoked after adding additional patterns
+	 * One can use this method to add additional classpath resource patterns to scan, if necessary Note,
+	 * that one will need to ensure that the refreshCache() method is invoked after adding additional
+	 * patterns
 	 */
 	public void addClasspathPatternToScan(String classpathPatternToScan) {
 		classpathPatternsToScan.add(classpathPatternToScan);
 	}
-
+	
 	/**
 	 * One can use this method to add additional classpath initializer domains to scan, if necessary
-	 * Note, that one will need to ensure that the refreshCache() method is invoked after adding additional domains
+	 * Note, that one will need to ensure that the refreshCache() method is invoked after adding
+	 * additional domains
 	 */
 	public List<String> getDomainsToScan() {
 		return domainsToScan;
@@ -365,12 +366,12 @@ public class InitializerMessageSource extends AbstractMessageSource implements M
 	public Map<String, String> getFallbackLanguages() {
 		return fallbackLanguages;
 	}
-
+	
 	/**
-	 * One can use this method to add a fallback language for a particular language
-	 * For example, to indicae that Haitian Kreyol (ht) should fallback to French prior to falling back to the
-	 * System locale, you would use this method to indicate that.
-	 * This does not require refreshCache() to be called, but will take immediate effect.
+	 * One can use this method to add a fallback language for a particular language For example, to
+	 * indicae that Haitian Kreyol (ht) should fallback to French prior to falling back to the System
+	 * locale, you would use this method to indicate that. This does not require refreshCache() to be
+	 * called, but will take immediate effect.
 	 */
 	public void addFallbackLanguage(String language, String fallbackLanguage) {
 		fallbackLanguages.put(language, fallbackLanguage);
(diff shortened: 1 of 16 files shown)


Ruhanga (commented):
@mseaton, thanks a bunch for raising this PR.  Would you also add a `readme` note highlighting this enhancement.

api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
```diff
@@ -50,104 +32,227 @@
...
-				message = message.replace("}}", "'}}'");
-				pmm.put(code, new PresentationMessage(code, locale, message, null));
+		}
+		return ret;
+	}
+	
+	/**
+	 * @return an array of message property resource file names found on the classpath
+	 */
+	protected Map<String, Properties> getMessagePropertyResourcesFromClasspath() {
```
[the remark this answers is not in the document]
mseaton: @mks-d no, the details of how this is done in core are buried in private methods within the core message source (org.openmrs.messagesource.impl.MutableResourceBundleMessageSource).  I tried very hard to build on this core message source or, failing that, to build on the ReloadableResourceBundleMessageSource that the core message source extends and which is provided by Spring.  In the end, I was unable to get all of my unit tests to pass using these techniques and came to the conclusion that it was not possible.

In the existing code, the InitializerMessageSource adds the core message source as a parent, which ends up delegating to that message source if the InitializerMessageSource does not resolve a message.  The way I understand this to work is that:

a) The InitializerMessageSource will look for matches for a code in a Locale.  If this returns null, then
b) The parent will be delegated to, to look for matches for a code in a Locale

Now, the way the parent works (from my understanding) is that it is given an ordered array of resource bundle locations to check (i.e. a directory that has message property files it it for various locales).  It will check each in turn and if it finds a matching message, it will return it, without checking later resources in the array.  What this means is that if you are looking for a message for "fr_FR", and it finds a message for "fr" in the resource bundle at array index 1, it will return it, even if a better matching message is defined in "fr_FR" in the resource bundle at array index 2.  It's not really possible to ask the parent for the message at a specific locale and work with this.  You can only ask the parent to resolve a message a given locale and have it return a match at this (and less specific) locales.

Ultimately, I think there are issues like this that I would consider bugs in both the Initializer Message Source and _also_ in the OpenMRS core message source, and I wanted to address all of these.  So, in this new Message Source, I am basically implementing what I think a new OpenMRS message source should look like, that would handle all of the core needs and allow extension by Initializer.  Ultimately if this is successful and after an iteration or two of additional features, etc. I would encourage us to harvest this back into core as a replacement.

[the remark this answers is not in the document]
mseaton: Beyond this, there are things that I am doing in this message source that core is not, and likely should be, around which specific message properties files are loaded from the classpath.  I suspect that core is loading all of the message properties that are in module dependencies as well (including multiple versions of core).  I think if we were to write a collection of unit tests for just the core message source, with expected behaviors around how module messages should override core messages, or each other, and how these should handle messages with variant, country, and language options, or with fallback locales, then we will find that there are a number of these tests that will fail with core.

mseaton:
> @mseaton, thanks a bunch for raising this PR. Would you also add a `readme` note highlighting this enhancement.

@Ruhanga - yes, if helpful.  It wasn't 100% clear to me what this should look like.  I'd consider this more of a bug fix than an enhancement at this stage.  @mks-d ?

mogoodrich:
@mseaton haven't had a chance to review detail yet, will look at this tomorrow, but re: "namely that one should be able to configure the "ht" Locale to use "fr" as a fallback language before finally falling back to the system locale"...  I think that was originally working with the main OpenMRS Message Source by setting the GP "default_locale" to "fr".  It sounds like what you have adds is more flexible so worth keeping, but we might want to include the "default_locale" GP in the fallback chain (if you haven't already), ie after the language-specific fallback but before the system locale fallback.

mseaton:
> I think that was originally working with the main OpenMRS Message Source

As I wrote above, I do think this is a nice feature to add in @mogoodrich , but I haven't yet found any evidence that OpenMRS does this in the core message source.  I'd be very interested if you could reproduce this or point me to code that is in place that does this.

Note, if we were to implement this here, the static OpenMRS methods: LocaleUtility.getDefaultLocale() and LocaleUtility.getLocalesInOrder() are likely useful.

mogoodrich:
Interesting @mseaton ... I haven't looked at the code, but I thought this was working at some point, and was how we had fallback from Kreyol to French working (French is our default language in Haiti).  However, as this fallback no longer seems to be working, it's entirely possible that this was a feature in an earlier version of OpenMRS that was broken or removed at some point.  If your exploratio shows that the current message source doesn't do this, I'm fine with skipping.

api/src/test/java/org/openmrs/module/initializer/InitializerMessageSourceIntegrationTest.java
```diff
@@ -9,79 +9,128 @@
...
-		// setup
-		int initSize = inizSrc.getPresentations().size();
-		MessageSourceService coreSrc = (MessageSourceService) applicationContext.getBean("messageSourceServiceTarget");
-		coreSrc.addPresentation(new PresentationMessage("helloWorld.692af26f1c07", Locale.ENGLISH, "Hello World!", null));
-		coreSrc.addPresentation(
-		    new PresentationMessage("helloWorld.692af26f1c07", Locale.FRENCH, "Bonjour le Monde !", null));
+	public void shouldLoadTranslationsWithAppropriateFallbacks() {
+		Context.setLocale(HAITIAN_KREYOL);
+		testMessage("Only In ht", "messageOnlyInHt"); // Direct match
+		testMessage("messageOnlyInFrFr", "messageOnlyInFrFr"); // Fallback to message code
```
mogoodrich, line 92: Super nit, but comment should probably say something like "no match, fallback to message code"... I was slightly confused at first

api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
```diff
@@ -50,104 +32,227 @@
...
-	 * @return the cached messages, for this source only
+	 * This method is called automatically by Spring when this Bean is instantiated in the context See
+	 * moduleApplicationContext.xml
+	 */
+	public void initialize() {
+		setUseCodeAsDefaultMessage(true);
+		addClasspathPatternToScan("classpath*:messages*.properties");
+		addDomainToScan(InitializerConstants.DOMAIN_ADDR);
+		addDomainToScan(InitializerConstants.DOMAIN_MSGPROP);
+		addFallbackLanguage("ht", "fr");
```
mogoodrich, line 111: Wait, is the idea that we are hardcoding these fallbacks into the message source itself? That doesn't seem correct.

api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
```diff
@@ -50,104 +32,227 @@
...
+ * will take first precedence, followed by those in modules, and finally by those in OpenMRS core.
+ * Ultimately, a Map backs this source, with Locale and message code as keys to the Map, and thus
+ * values loaded later will replace those loaded earlier in the process. The requested Locale always
+ * takes precedence over the order of sources. For example, if a message is requested for Locale
+ * 'fr_FR', all sources are first searched for a match to 'fr_FR' prior to then searching for 'fr'
+ * across all sources. If no match is found in a given Locale, then the fallback is to search for
+ * the best match in the System Locale. If no match is found in the System Locale, then the message
+ * code itself is returned as a final fallback. This source allows for supporting additional
+ * fallback languages, as well as defining additional classpath patterns and domains to search for
+ * message property files. See method javadoc for further details.
```
mogoodrich, line 66: I would reference the method name here, ie "addFallbackLanguage"

mogoodrich (changes requested):
Generally looks good, one small comment regarding hardcoding the Kreyol fallback (probably something you forgot to clean up)?

api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
mseaton: My idea is that we would provide a means to add fallback languages, but that yes, we would specifically make a decision that - by default - French is the fallback for Haitian Kreyol for message localization.  This feels right to me.

api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
mseaton: Done

api/src/test/java/org/openmrs/module/initializer/InitializerMessageSourceIntegrationTest.java
mseaton: Done

api/src/main/java/org/openmrs/module/initializer/InitializerMessageSource.java
mseaton: OK, talked this over with @mogoodrich and agreed to remove this hard-coded fallback in favor of letting this be configured as-needed.

mseaton:
@mks-d - FYI, I have just created a copy of this class and added it into one of our PIH modules, and wired it up as the main message source.  This will give us some opportunities to test this out in the wild with our actual distribution running across various countries and test servers, and provide better information as to how this works beyond any code review that I fear will not otherwise allow us to reach the level of confidence needed to merge this in.  Let's circle back around this in a few days and see how things stand and then aim to get this merged in if tests are successful?  Please review in the meantime and let me know if anything looks concerning.

mks-d (commented):
>I have just created a copy of this class and added it into one of our PIH modules, and wired it up as the main message source. This will give us some opportunities to test this out in the wild with our actual distribution running across various countries and test servers

@mseaton I wonder if your in-house testing on the copy of this enhanced class are positive? If yes I'd be keen to merge this in.

mseaton:
> > I have just created a copy of this class and added it into one of our PIH modules, and wired it up as the main message source. This will give us some opportunities to test this out in the wild with our actual distribution running across various countries and test servers
> 
> @mseaton I wonder if your in-house testing on the copy of this enhanced class are positive? If yes I'd be keen to merge this in.

@mks-d yes they have been positive.  I've just updated the PR to more accurately reflect our in-house version, which had a few changes made to it since the initial copy.  It was pretty much unchanged, with the exception being how it is setup initially.  I found that just relying on the ApplicationContextAware handling was not reliable enough in terms of ordering.  Instead, I changed it to be a component and then explicitly set it as the active message source in the started method of the activator.  This ensures that it is set after all other earlier modules and core have loaded and that this is done last.

I'd recommend we merge this in and provide an opportunity early enough prior to the next release to allow existing users to test it out, but overall we have been happy with this.

mks-d (approved):

Issue #154, opened 2021-11-04
DisplaysCsvParser causes all rows with retired/voided data to fail with an exception


mseaton:
The implementation of the DisplaysCsvParser that was recently introduced in #95 has the following implementation in its boostrap(CsvLine) method ([here](https://github.com/mekomsolutions/openmrs-module-initializer/blob/81756c905ca27b10e643b88411628a1a486af778/api/src/main/java/org/openmrs/module/initializer/api/display/DisplaysCsvParser.java#L40-L43)):

```java
public OpenmrsObject bootstrap(CsvLine line) throws IllegalArgumentException {
    if (BaseLineProcessor.getVoidOrRetire(line)) {
        throw new APIException("A voided or retired object cannot be internationalized. "
                + "Check the implementation of this parser: " + getClass().getCanonicalName());
    }
    
    return bootstrapParser.bootstrap(line);
}
```

The result of this is that for any CSV domains that contain rows with Voided or Retired data, the pre-loader will fail for these rows by throwing an Exception.

Now, when Initializer is configured to ignore exceptions that occur on a given row (which is the default, out of the box behavior), this does not result in any real issues, as failures will happen during pre-loading, and processing will happily continue and the actual row loading will work.  And since this is how every unit test in the module is configured to run as well, this issue did not present itself during development.   Still, these row issues are reported and you can see them if you inspect row errors, which is not ideal, since you really want these row errors to consist of actual problems that need to be dealt with, not background non-issues.

However, more concerningly, once you start running Initializer with stricter loading rules, this starts to fail.  Indeed, this new version of Initializer fails to load any of our distributions at PIH, which all have voided / retired rows and which we load in "strict" mode.

The issues can be reproduced by either:

a) Using loader.loadUnsafe(null, true) - i.e. in a unit test, as an alternative to loader.load()

-or-

b) Setting the runtime property:  `initializer.startup.load = fail_on_error`

For example, go into EncounterTypesLoaderIntegrationTest#load_shouldLoadEncounterTypesAccordingToCsvFiles and change the first ine from `loader.load()` to `loader.loadUnsafe(null, true)` and you will see this unit test fail.  And running it as-is will also show a row error in the log, even though the unit test passes.

[...]

PR #158, opened 2021-11-08
[#154]: DisplaysCsvParser failures to not cause subsequent loading to fail in strict mode by default.

Ruhanga:


Ruhanga (commented):
cc @mseaton, @mks-d

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseFileLoader.java
```diff
@@ -62,13 +62,25 @@ private File preload(final File file, boolean doThrow) {
...
+	}
+	
 	@Override
 	public void loadUnsafe(List<String> wildcardExclusions, boolean doThrow) throws Exception {
 		
 		final ConfigDirUtil dirUtil = getDirUtil();
 		
 		dirUtil.getFiles(getFileExtension(), wildcardExclusions).stream().map(f -> toOrderedFile(f)).sorted()
-		        .map(f -> preload(f, doThrow)).filter(f -> !dirUtil.getChecksumIfChanged(f).isEmpty()).forEach(file -> {
+		        .map(f -> preload(f, failOnPreLoadError(false))).filter(f -> !dirUtil.getChecksumIfChanged(f).isEmpty())
```
mks-d, line 82: This should pass the actual value to the function: `failOnPreLoadError(doThrow)`.

Also the idea would that loaders would expose two methods about their potential preloader capabilities:
1. `isPreloader()` - that says wether the loader is a preloader.
2. `isThrowingPreloader(bool)` - that says whether the preloader throws or not.

Let's add `isThrowingPreloader(bool)` to `Loader` as well, [here](https://github.com/mekomsolutions/openmrs-module-initializer/blob/81756c905ca27b10e643b88411628a1a486af778/api/src/main/java/org/openmrs/module/initializer/api/loaders/Loader.java#L15).

mks-d (changes requested):

api/src/test/java/org/openmrs/module/initializer/api/loaders/BaseFileLoaderTest.java
```diff
@@ -112,4 +112,114 @@ public void loadUnsafe_shouldThrowEarly() throws Exception {
 		Assert.assertTrue(thrown.getMessage().endsWith("Error right from file 1."));
 		verify(dirUtil, times(0)).writeChecksum(any(), any());
 	}
+	
+	@Test
+	public void loadUnsafe_shouldThrowOnPreloadAsDefaultGivenDoThrowTrue() throws Exception {
+		// setup
+		BaseFileLoader fl = new BaseFileLoader() {
+			
+			TestLoader test = new TestLoader();
```
mks-d, line 121: Since `TestLoader` is used twice, make it a nested class of `BaseFileLoaderTest`, and reuse it twice.

README.md
```diff
@@ -147,7 +147,7 @@ https://github.com/mekomsolutions/openmrs-module-initializer/issues
...
 * Backward-compatible overhaul of the concepts domain that provides full flexibility for managing concept names.
 * Bulk loading of metadata entities i18n display messages specified under `display:xy` headers.
-* (_For devs._) Introduced a pre-loading mechanism to `BaseFileLoader` that allows checksums-independent loading of transient information out of the config files before the actual metadata are loaded.
+* (_For devs._) Introduced a pre-loading mechanism to `BaseFileLoader` that allows checksums-independent loading of transient information out of the config files before the actual metadata are loaded. The behavior of all sub classes not overriding `throwingOnPreload` to return false, is to always, as the default, throw exceptions early and intercept the actual loading of the files given the runtime setting of `initializer.startup.load = fail_on_error` was configured.
```
mks-d, line 150: >Introduced a pre-loading mechanism to `BaseFileLoader` that allows checksums-independent loading of transient information out of the config files through a pre-loader, before the actual metadata are loaded. Each loader controls whether its pre-loader throws on error or is allowed to fail. By default pre-loaders are allowed to fail.

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseFileLoader.java
```diff
@@ -62,13 +62,26 @@ private File preload(final File file, boolean doThrow) {
...
 	}
 	
+	/**
+	 * During pre-loading of a domain file, an exception may occur preventing its actual loading in
+	 * strict mode(i.e doThrow is true). Loaders that do not have to fail loading given an exception
+	 * occurred when pre-loading would have to override this function to return false.
+	 * 
+	 * @param doThrow whether to throw exception on not
+	 * @return true or false.
+	 */
```
mks-d, line 72:
```java
/**
 * If the loader is fitted with a pre-loader, this method provides control on whether the pre-loader should throw on error or is allowed to fail.
 * 
 * @param doThrow Says whether Initializer is configured to throw on error or not.
 * @return true is the loader's pre-loader should throw on error, false if the loader's pre-loader is allowed to fail.
 */
```

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseCsvLoader.java
```diff
@@ -72,6 +72,15 @@ protected void preload(final File file) throws Exception {
 		displaysLoader.load(file);
 	}
 	
+	/**
+	 * By default all CSV loaders do not allow pre-loaders to throw exceptions in their pre-loading
+	 * phase.
+	 */
```
mks-d, line 78:
```java
/**
 * By default CSV loaders do not allow their pre-loader to throw.
 */
```

mks-d (changes requested):
Thanks @Ruhanga, a couple of rewrites of the README and JavaDocs. Plus some minor changes.

Ruhanga (commented):
Thanks for the review @mks-d. I've addressed the requested changes. Could you review again?

mks-d (approved):

Issue #165, opened 2021-12-10
Idgen domain needs to support remote sources and identifier pools


mseaton:
The idgen domain is currently limited such that it can only create SequentialIdentifierGenerator sources.  This domain should be expanded such that both RemoteIdentifierSource and IdentifierPool sources can also be created/edited/maintained with the domain.

PR #166, opened 2021-12-10
[#165] Idgen domain needs to support remote sources and identifier pools

mseaton:


api/src/main/java/org/openmrs/module/initializer/api/idgen/RemoteIdentifierSourceLineProcessor.java
```diff
@@ -27,12 +28,36 @@ public IdgenSourceWrapper fill(IdgenSourceWrapper instance, CsvLine line) throws
...
+	/**
+	 * Remote Identifier Sources have properties that contain sensitive information
+	 * For this reason, we enable the ability to specify that these properties should
+	 * be read from runtime or system properties, rather than fixed values in the CSV
+	 */
+	protected String getRequiredProperty(CsvLine line, String header) {
+		String val = line.get(header, true);
+		if (val == null) {
+			throw new IllegalArgumentException(header + " is required");
+		}
```
mks-d, line 49: `line.get(header, true)` would [throw](https://github.com/mekomsolutions/openmrs-module-initializer/blob/5098c1058cb710ce683dde8dca7aba877e0f2ef8/api/src/main/java/org/openmrs/module/initializer/api/CsvLine.java#L50) if the header wasn't found no?

Don't you precisely want to do the contrary: use `line.get(header, false)` and if the value is `null` proceed through to the props?

api/src/main/java/org/openmrs/module/initializer/api/idgen/RemoteIdentifierSourceLineProcessor.java
```diff
@@ -27,12 +28,36 @@ public IdgenSourceWrapper fill(IdgenSourceWrapper instance, CsvLine line) throws
...
+		if (val.toLowerCase().startsWith("property:")) {
+			String property = val.substring(9);
+			if (System.getProperties().containsKey(property)) {
+				val = System.getProperty(property);
+			} else {
+				val = Context.getRuntimeProperties().getProperty(property);
+			}
+			if (val == null) {
+				throw new IllegalArgumentException(header + " is required but property " + property + " is not found");
```
mks-d, line 58: To make the logs slightly more readable:
```
header + " is required but property '" + property + "' is not found."
```

api/src/main/java/org/openmrs/module/initializer/api/idgen/RemoteIdentifierSourceLineProcessor.java
```diff
@@ -27,12 +28,36 @@ public IdgenSourceWrapper fill(IdgenSourceWrapper instance, CsvLine line) throws
...
+	 * Remote Identifier Sources have properties that contain sensitive information
+	 * For this reason, we enable the ability to specify that these properties should
+	 * be read from runtime or system properties, rather than fixed values in the CSV
+	 */
+	protected String getRequiredProperty(CsvLine line, String header) {
+		String val = line.get(header, true);
+		if (val == null) {
+			throw new IllegalArgumentException(header + " is required");
+		}
+		if (val.toLowerCase().startsWith("property:")) {
```
mks-d, line 50: Could we choose from either of
* `"application property:"`
* `"system property:"`
* `"runtime property:"` 

?
Or you think that none are appropriate?

If there is an appropriate one then this would lead to renaming the method as well to, for example, `getApplicationProperty`.

api/src/main/java/org/openmrs/module/initializer/api/idgen/RemoteIdentifierSourceLineProcessor.java
```diff
@@ -27,12 +28,36 @@ public IdgenSourceWrapper fill(IdgenSourceWrapper instance, CsvLine line) throws
...
 		instance.setIdentifierSource(source);
 		
 		return instance;
 	}
+
+	/**
+	 * Remote Identifier Sources have properties that contain sensitive information
+	 * For this reason, we enable the ability to specify that these properties should
+	 * be read from runtime or system properties, rather than fixed values in the CSV
+	 */
```
mks-d, line 44: This is nice meat for the README file, as for the JavaDoc it should just says what it does:

>Attempts to read the value from the CSV line as usual or falls back to reading the specified system or runtime property. Eg "property:foo" will attempt to read first the system property named "foo" or eventually the runtime property named "foo".

mseaton:
I've addressed a number of your comments and added responses to the rest @mks-d .  Let me know thoughts on getting this merged.

api/src/main/java/org/openmrs/module/initializer/api/idgen/RemoteIdentifierSourceLineProcessor.java
ibacher:
```suggestion
				throw new IllegalArgumentException(header + " is required but property '" + property + "' is not found");
```

api/src/main/java/org/openmrs/module/initializer/api/idgen/RemoteIdentifierSourceLineProcessor.java
ibacher: @mks-d As I read it, the intention is to have `property:<some_property_name>` in the line and replace that with the value gotten by looking up the property.

ibacher (approved):
@mseaton Thanks for this. It looks good to me. Just need to confirm with the validator.

ibacher (changes requested):
Need to change things so the validator works, i.e., the domain can really be executed.

mseaton:
@ibacher - does the validator work for the idgen domain when run by itself _prior_ to this commit?  This commit doesn't introduce the domain, and I'm curious if this error was pre-existing...

Ruhanga:
>does the validator work for the idgen domain when run by itself
I'd think no @mseaton, because of some meta-data the domain depends on that is not yet in the database for example the `Identifier type`. For this you'd want to load all the metadata necessary for the domain to load successfully.

ibacher:
@mseaton @Ruhanga Yeah, that's fair. I was hoping to quickly verify that everything was ok and didn't have much of an opportunity to dig into why it failed. Looking into that now.

ibacher:
I get a very similar set of errors if I don't restrict the domain, but then again, they also appear on `master` so I'm assuming it's not the fault of this PR per se.

mseaton:
Thanks @ibacher - so would that imply that we don't need to fix that validator issue to merge this PR?  Since the issue in question already exists and we can ticket independently?

ibacher:
@mseaton Yes, that's the case. I just want to do a test on my end and then I think this can be merged.

Issue #136, opened 2021-09-08
Concept domain should support broader configuration of concept names


mseaton:
In our Concept dictionary, we have situations where:

* We have concept names with a null concept_name_type (these are typically used for synonyms)
* We have more than one name of a given type on a concept (typically multiple synonyms)
* We have concepts where the one that is locale_preferred in a given language is not the fully specified name, but rather a synonym or a short name
* We have concept names for which we may want to be able to explicitly set the uuid

Proposed solution:

Support a new set of name headers (the existing name headers would also continue to be supported for backwards compatibility and for simpler cases).  These headers would be identified by specific "Header Prefixes" (the part before the <N>) and would communicate to the loader that these should be treated as components of a concept name.  The <N> would be any arbitrary suffix that the implementer wants to use, as long as it's unique per name, and shared across each name component.  

Name <N>
Name Locale <N> (optional)
Name Locale Preferred <N> (optional)
Name Type <N> (optional)
Name Uuid <N> (optional)

* If the Local Preferred column is not included, or if it is not populated for a given Concept, then the first Concept found on the row in a given locale will be considered to be the one that should be preferred.  So if a synonym should be marked as preferred, it should be the first one processed in it's locale in a given concept row.  See example for how the Metadata Deploy module did this here:  https://github.com/openmrs/openmrs-module-metadatadeploy/blob/master/api/src/main/java/org/openmrs/module/metadatadeploy/builder/ConceptBuilder.java#L63

*If the Name Type column is not included, or if it is not populated for a given Concept, then the first Concept found in a given locale will be considered the fully specified name, and any additional concepts found will be considered synonyms

* If Name Locale is not included, or if it is not populated for a given Concept, then Context.getLocale() will be used and will set the name to the default system locale.

@mks-d interested in your thoughts.  @mogoodrich FYI

mks-d:
@mseaton I'm sure you have pondered this and concluded as per your description, but wasn't it possible to drop the `Name Type` column altogether and have three cases:
1. `Fully specified name *` (those headers would always be for FSNs)
2. `Short name *` (those headers would always be for short names)
3. `Name {n} *` (those headers would always be for synonyms)

?

Also isn't it easier to implement if `Name {n}` is always the prefix, like this:
`Name {n}`
`Name {n} Locale` (optional)
`Name {n} Locale Preferred` (optional)
`Name {n} UUID` (optional)
?

[...]

PR #167, opened 2021-12-13
[#136] Concept names defined with no UUIDs to recreate "same" existing names with unmatched UUIDs.

Ruhanga:
Before `ConceptName`s were made more configurable [before #136], existing `ConceptName`s were detached/removed from the main `Concept` in favour of new names brought in by the config irrespective of whether the string names were equal or not. This behaviour should be maintained.

Currently CSV names not configured with a uuid could match existing names with different `uuid`s and not get their uuids updated with the default seeded uuid of their associated concepts. 

This PR therefore allow for the CSV concept names, not configured with a uuid to be saved with a default seeded uuid from the associated concept. This may resulting in the voiding of existing equal names but with different `uuid`s in favour of the CSV defined name.


api/src/test/java/org/openmrs/module/initializer/api/ConceptsLoaderIntegrationTest.java
```diff
@@ -334,6 +334,26 @@ public void load_shouldLoadConceptsAccordingToCsvFiles() {
...
+		loader.load();
+		
+		// verify
+		Assert.assertNotEquals(originalConceptNameUuid,
+		    cs.getConceptByUuid(existingConceptUuid).getName(localeEn).getUuid());
+		Assert.assertEquals(
+		    Utils.generateUuidFromObjects(existingConceptUuid, nameString, ConceptNameType.FULLY_SPECIFIED, Locale.ENGLISH),
+		    cs.getConceptByUuid(existingConceptUuid).getName(localeEn).getUuid());
+	}
+
```
mseaton, line 356: I understand the goal here and the desire to do this, but I think we need to be very careful and I have some hesitation around this.  There are a lot of inner workings of OpenMRS, particularly in modules like Sync, that make assumptions about uuids being non-changing.  We might be better off voiding and recreating with a new locale, though that also has some understandable downsides.

Ruhanga: I'd go for the later suggestion with an addition removal of the name from the concept names. We could then have an additional concept name processor that'd provide a way to bootstrap an independently existing name to associate it with a concept if it's desired to reuse it.

Ruhanga:
@mseaton, @mks-d, after a conversation with @ibacher and following the suggestion you mentioned,  I've added commits resulting in existing names having different uuids from equivalent csv names being voided with a modification on the unit tests to this effect.

api/src/test/resources/testAppDataDir/configuration/concepts/concepts_names.csv
```diff
@@ -2,4 +2,4 @@ Uuid,Fully specified name:en:UUID,Fully specified name:en,Fully specified name:e
 58083852-303f-11ec-8d2b-0242ac110002,e91ab3ad-303f-11ec-8d2b-0242ac110002,Red,Rojo,R,R,,Maroon,,,,Misc,N/A,,
 5dcaf167-303f-11ec-8d2b-0242ac110002,fe9c8c03-303f-11ec-8d2b-0242ac110002,Blue,Azul,B,A,,Navy,TRUE,Azulado,Baby Blue,Misc,N/A,,
 61214827-303f-11ec-8d2b-0242ac110002,,Green,Verde,,,,,,,,Misc,N/A,,
-4cfe07b0-3061-11ec-8d2b-0242ac110002,,Lemon Yellow,,Y,,c0c238b3-3061-11ec-8d2b-0242ac110002,Yellow,TRUE,,,Misc,N/A,,
+4cfe07b0-3061-11ec-8d2b-0242ac110002,,Lemon Yellow,,Y,,c0c238b3-3061-11ec-8d2b-0242ac110002,Yellow,TRUE,,,Misc,N/A,,
```
mks-d, line 5: @Ruhanga I think you could untouch that file right?

api/src/test/java/org/openmrs/module/initializer/api/ConceptsLoaderIntegrationTest.java
```diff
@@ -385,12 +385,17 @@ public void load_shouldLoadConceptNamesAccordingToCsvFiles() {
 		
 		// Green is an existing Concept
 		// It has two existing names, and no uuids are specified in the CSV.
-		// Since the name, type, and locale are the same in the CSV as in the DB, it should match, and not recreate.
+		// Even if the name, type, and locale are the same in the CSV as in the DB, it should not match
+		// given uuids differ but recreate new names with seeded uuids.
```
mks-d, line 389: @Ruhanga could you attempt to rewrite this comment?

Something along the lines that the 'green' concept has 2 concept names prior to loading (FSN "green" and FSN "verde") and will end up with 4 concept names after having loaded (FSN "green" again, FSN "verde" again but with different UUIDs, and explain why they are different, and voided FSN "green" and voided FSN "verde"). In short, let's explain in more details what happens because it's tricky and unclear (to me at least).

mks-d (commented):
@Ruhanga also maybe rewrite the headline?
>CSV concept names to be recreated, voiding existing ones with similar names given they donot share same uuid

Similar names? What do you mean by "similar"?

mks-d (approved):

mks-d:
My description of this change is the following: when a CSV does not set UUIDs for the concept names that it defines **then it implicitly says that the UUIDs should be _seeded_**. Therefore any existing names that match (same type, same locale, same name string) will be voided if their UUID is in contravention with the UUID specified by the CSV (that is: seeded UUID).

So yes maybe there are existing names whose UUID was once set to be something specific on purpose (and maybe other processes rely on those UUIDs to be what they are, like a sync process or otherwise) in which case the CSV redefinition of those names need to respecify those UUIDs again, otherwise they are at risk of being voided.

mseaton:
@mks-d and @Ruhanga - sorry I was offline for a bit.  This all sounds good.  Thanks for the work on this.

icrc-fdeniger:
Hi,
I've just tested the last version 2.3.0-SNAPHOST, and I got a lot of error like:
`org.openmrs.api.DuplicateConceptNameException: 'XXXXX' is a duplicate name in locale 'en' for the same concept`

Before using the last version, we had 195k concept_names with 2.5k voided concept_names
After restarting OpenMRS with the last iniz version, we got 7.5k voided concept names. I'd expect to have almost 195k voided concept_name.

Ruhanga:
Thanks @icrc-fdeniger for reporting this. It is possible that the new `ConceptNames`s are saved before the old ones are updated which results into the error you have shared. To resolve this I've pushed a [PR](https://github.com/mekomsolutions/openmrs-module-initializer/pull/170) updating existing names before new ones are created/added.

PR #170, opened 2022-01-03
[#136] Existing concept names to be resaved before adding new ones.

Ruhanga:
issue: #136

After testing the latest Iniz snapshot on an implementation with multiple concepts being reloaded, random `Concept`s fail to load because their existing names are not yet updated by the time new names are being saved. This happens especially when the existing name(s) are being voided and new names added. 

Consider a `Concept` with 3 names `Green`, `Yellow` and `Red` in a single locale being reloaded and having random uuids.
The 3 existing names should be voided and 3 new names recreated having seeded uuids.
Since saving and/or updating concept names happens randomly, it is possible that the new name Yellow with a seeded uuid gets saved before the existing name Yellow with a random uuid is updated as voided. In which case a `DuplicateConceptNameException` error occurs resulting in the overall failure to load the concept. Checkout the [ConceptValidator](https://github.com/openmrs/openmrs-core/blob/2.1.0/api/src/main/java/org/openmrs/validator/ConceptValidator.java#L175).

This PR is therefore a patch up on the previous work done on #136 ensuring existing concept names are updated before new ones are added.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
@@ -97,8 +97,7 @@ public class ConceptsCsvParser extends CsvParser<Concept, BaseLineProcessor<Conc
 		        .collect(Collectors.toList());
 		List<ConceptName> allConceptNames = instance.getNames(true).stream().collect(Collectors.toList());
 		
-		if ((newConceptNames.containsAll(allConceptNames) && allConceptNames.containsAll(newConceptNames))
-		        && (newConceptNames.size() == allConceptNames.size())) {
+		if (newConceptNames.containsAll(allConceptNames) && (newConceptNames.size() == allConceptNames.size())) {
 			return conceptService.saveConcept(instance);
 		} else {
 			// First update existing names before saving new ones
--- a/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
@@ -1,6 +1,7 @@
 package org.openmrs.module.initializer.api.c;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Locale;
 import java.util.stream.Collectors;
@@ -94,20 +95,18 @@ public class ConceptsCsvParser extends CsvParser<Concept, BaseLineProcessor<Conc
 	@Override
 	public Concept save(Concept instance) {
 		List<ConceptName> newConceptNames = instance.getNames(true).stream().filter(cn -> cn.getId() == null)
-		        .collect(Collectors.toList());
-		List<ConceptName> allConceptNames = instance.getNames(true).stream().collect(Collectors.toList());
+		        .sorted(Comparator.comparing(ConceptName::getName)).collect(Collectors.toList());
 		
-		if (newConceptNames.containsAll(allConceptNames) && (newConceptNames.size() == allConceptNames.size())) {
-			return conceptService.saveConcept(instance);
-		} else {
+		if (!newConceptNames.equals(instance.getNames(true).stream().sorted(Comparator.comparing(ConceptName::getName))
+		        .collect(Collectors.toList()))) {
 			// First update existing names before saving new ones
 			// This is to prevent possible DuplicateConceptNameException because ConceptName comparison happens with names existing in the database
 			// See https://github.com/openmrs/openmrs-core/blob/2.1.0/api/src/main/java/org/openmrs/validator/ConceptValidator.java#L175
 			instance.getNames(true).removeAll(newConceptNames);
 			instance = conceptService.saveConcept(instance);
 			instance.getNames(true).addAll(newConceptNames);
-			return conceptService.saveConcept(instance);
 		}
+		return conceptService.saveConcept(instance);
 		
 	}
 	
--- a/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
@@ -3,6 +3,7 @@ package org.openmrs.module.initializer.api.c;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.openmrs.Concept;
@@ -92,13 +93,12 @@ public class ConceptsCsvParser extends CsvParser<Concept, BaseLineProcessor<Conc
 	
 	@Override
 	public Concept save(Concept instance) {
-		List<ConceptName> newConceptNames = new ArrayList<ConceptName>();
-		for (ConceptName name : instance.getNames(true)) {
-			if (name.getId() == null) {
-				newConceptNames.add(name);
-			}
-		}
-		if (newConceptNames.size() == instance.getNames(true).size()) {
+		List<ConceptName> newConceptNames = instance.getNames(true).stream().filter(cn -> cn.getId() == null)
+		        .collect(Collectors.toList());
+		List<ConceptName> allConceptNames = instance.getNames(true).stream().collect(Collectors.toList());
+		
+		if ((newConceptNames.containsAll(allConceptNames) && allConceptNames.containsAll(newConceptNames))
+		        && (newConceptNames.size() == allConceptNames.size())) {
 			return conceptService.saveConcept(instance);
 		} else {
 			// First update existing names before saving new ones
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
```diff
@@ -89,7 +92,24 @@ public Concept bootstrap(CsvLine line) throws IllegalArgumentException {
 	
 	@Override
 	public Concept save(Concept instance) {
-		return conceptService.saveConcept(instance);
+		List<ConceptName> newConceptNames = new ArrayList<ConceptName>();
+		for (ConceptName name : instance.getNames(true)) {
+			if (name.getId() == null) {
+				newConceptNames.add(name);
+			}
+		}
```
mks-d, line 100: Looks like streams were made to one-line this kind of stuff: filtering stuff out of a collection into another collection.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
```diff
@@ -89,7 +92,24 @@ public Concept bootstrap(CsvLine line) throws IllegalArgumentException {
...
+		if (newConceptNames.size() == instance.getNames(true).size()) {
```
mks-d, line 101: This is not ok right, you want to validate that those are the two same collections, not just that they have the same size.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
```diff
@@ -89,7 +93,23 @@ public Concept bootstrap(CsvLine line) throws IllegalArgumentException {
 	
 	@Override
 	public Concept save(Concept instance) {
-		return conceptService.saveConcept(instance);
+		List<ConceptName> newConceptNames = instance.getNames(true).stream().filter(cn -> cn.getId() == null)
+		        .collect(Collectors.toList());
+		List<ConceptName> allConceptNames = instance.getNames(true).stream().collect(Collectors.toList());
+		
+		if ((newConceptNames.containsAll(allConceptNames) && allConceptNames.containsAll(newConceptNames))
+		        && (newConceptNames.size() == allConceptNames.size())) {
```
mks-d, line 101: I am sure there is better Java ways to check that two collections are identical. And actually this should work:
```java
newConceptNames.equals(allConceptNames)
```
In any case, since you check for the sizes to be equal then you don't need to use twice `containsAll`, this is enough:
```java
(newConceptNames.size() == allConceptNames.size()) && allConceptNames.containsAll(newConceptNames)
```

Ruhanga: I'd agree with `(newConceptNames.size() == allConceptNames.size()) && allConceptNames.containsAll(newConceptNames)` where as `newConceptNames.equals(allConceptNames)` wouldn't work well especially that I'm dealing with objects which in addition might be having differently ordered members.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
```diff
@@ -89,7 +93,22 @@ public Concept bootstrap(CsvLine line) throws IllegalArgumentException {
 	
 	@Override
 	public Concept save(Concept instance) {
-		return conceptService.saveConcept(instance);
+		List<ConceptName> newConceptNames = instance.getNames(true).stream().filter(cn -> cn.getId() == null)
+		        .collect(Collectors.toList());
+		List<ConceptName> allConceptNames = instance.getNames(true).stream().collect(Collectors.toList());
+		
+		if (newConceptNames.containsAll(allConceptNames) && (newConceptNames.size() == allConceptNames.size())) {
```
mks-d, line 100: Does this not work then?
```java
if (newConceptNames.equals(instance.getNames(true))) {
```
This would help not having to introduce the `allConceptNames` variable.

mks-d (commented):
We will need a test case.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
Ruhanga: `newConceptNames.equals(instance.getNames(true))` would work if both lists are actually ordered lists. However let me commit something similar to this.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptsCsvParser.java
```diff
@@ -89,7 +94,20 @@ public Concept bootstrap(CsvLine line) throws IllegalArgumentException {
...
+		        .collect(Collectors.toList()))) {
+			// First update existing names before saving new ones
+			// This is to prevent possible DuplicateConceptNameException because ConceptName comparison happens with names existing in the database
+			// See https://github.com/openmrs/openmrs-core/blob/2.1.0/api/src/main/java/org/openmrs/validator/ConceptValidator.java#L175
+			instance.getNames(true).removeAll(newConceptNames);
+			instance = conceptService.saveConcept(instance);
+			instance.getNames(true).addAll(newConceptNames);
+		}
 		return conceptService.saveConcept(instance);
+
```
ibacher, line 110: Does this actually work? `getNames()` looks like this:

```java
public Collection<ConceptName> getNames(boolean includeVoided) {
	if (names == null) {
		names = new HashSet<>();
	}

	return names.stream()
			.filter(n -> includeVoided || !n.getVoided())
			.collect(Collectors.toSet());
}
```

Which means that each call to `instance.getNames()` is returning a new `Set` instance, so, e.g., `getNames(true).removeAll()` and `getNames(true).addAll()` are operating on two different sets (and a third set is the one actually being persisted).

Seems like the implementation should be:

```java
Collection<ConceptName> finalConceptNames = instance.getNames(true);

Set<ConceptName> existingNames = finalConceptNames.stream().filter(cn -> cn.getId() != null).collect(Collectors.toSet());

// If we have modified any existing names these modifications should be applied before saving new ones
// This is to prevent possible `DuplicateConceptNameException`s that might occur if an existing concept
// name hasn't been voided
if (existingNames.size() > 0) {
	Set<ConceptName> newNames = finalConceptNames.stream().filter(cn -> cn.getId() == null).collect(Collectors.toSet());

	instance.setNames(existingNames);
	instance = conceptService.saveConcept(instance);

	Collection<ConceptName> updatedNames = instance.getNames(true);
	updatedNames.addAll(newNames);
	instance.setNames(updatedNames);
}

return conceptService.saveConcept(instance);
```

And, hey, we can skip all the sorting stuff. It might work to skip the `updatedNames` part and just do `instance.setNames(finalNames)`, but I'm less sure about that.

Ruhanga: Thanks @ibacher for your great review. I've had to go through a couple of debugging on an openmrs instance depending on this fix. And you were actually right, the concept configs go through without any expected error occurring.  I've had to also further play around with the openmrs [ConceptValidator](https://github.com/openmrs/openmrs-core/blob/2.1.x/api/src/main/java/org/openmrs/validator/ConceptValidator.java) when testing and confirmed that even updating the voided names before saving new ones won't work. See [here](https://github.com/openmrs/openmrs-core/blob/2.1.x/api/src/main/java/org/openmrs/validator/ConceptValidator.java#L56-L89). 

And also adding something like `instance.setNames(names);` results into the following error
`org.hibernate.HibernateException: A collection with cascade="all-delete-orphan" was no longer referenced by the owning entity instance: org.openmrs.Concept.names`

What I'm thinking might work is doing something like this below to bypass the validator, which might work as a hack though.

```java
Set<ConceptName> newNames = instance.getNames(true).stream().filter(cn -> cn.getId() == null)
        .collect(Collectors.toSet());
		
Set<ConceptName> existingNames = instance.getNames(true).stream().filter(cn -> cn.getId() != null)
        .collect(Collectors.toSet());
		
if (existingNames.size() > 0 && newNames.size() > 0) {
	for (ConceptName cn : newNames) {
				instance.removeName(cn);
	}
	Context.flushSession();
	for (ConceptName cn : existingNames) {
		instance.addName(cn);
	}
}
		
return conceptService.saveConcept(instance);
```

Thoughts @mks-d, @ibacher, @mseaton....

ibacher: Ah yes... I always forget about that issue with Hibernate's collections. Yes, your outline looks like a good way to go.

ibacher: Might need a `conceptService.saveConcept()` call before the `flushSession()` call.

mks-d (approved):
@ibacher merging this as this is becoming pressing.

@Ruhanga it would be good to see a test case added through a follow-up commit.

PR #172, opened 2022-01-13
[#136] Concept names to enforce their type before being set to their concept.

Ruhanga:
For some unexpected reason at runtime,  short names get added to their concepts as synonyms[`ConceptName.nametype=null`]. For names to maintain their type before they are saved/persisted in this PR, I've had to reset their types after they have been added to the concepts to prevent them from being saved as synonyms unexpectedly. This will in turn prevent concept validation errors that'd result later when saving the concepts.


Ruhanga (commented):
Hi @ibacher, sorry I think I missed your changes because we could have pushed changes simultaneously. Could you please still add your changes? Thanks.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptLineProcessor.java
```diff
@@ -159,7 +159,14 @@ else if (matchingNames.size() == 1) {
 			if (StringUtils.isEmpty(newName.getUuid())) {
 				newName.setUuid(generateConceptNameUuid(concept, newName));
 			}
+			// When adding names to concept, their name types may unexpectedly be changed to null/synonym
+			// which is undesired if its desired to maintain them.
+			// The name type is therefore reset after adding the name to the concept as a hack to maintain it 
+			// before it is saved later
```
mks-d, line 165: @Ruhanga have you identified where in Core this incorrect setting of the name type does occur?

ibacher:
@Ruhanga No problem... re-added.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptLineProcessor.java
ibacher: If I had to make a guess, it's at [`Concept#addName()`](https://github.com/openmrs/openmrs-core/blob/afc763bbd0206939cb1f4ad52506883e011688ea/api/src/main/java/org/openmrs/Concept.java#L1003), which seems to make a number of assumptions about the order that it's called in (i.e., the FSN must be the first name added; adding a FSN or short name to the concept for a locale converts the previous FSN or short name into a synonym...

Maybe outside the scope of this work, but we should make some effort to ensure that concept names are added in the expected order?

ibacher (approved):
Seems correct to me.

api/src/main/java/org/openmrs/module/initializer/api/c/ConceptLineProcessor.java
mseaton: @Ruhanga / @ibacher / @mks-d - this rings a bell - I ended up adding some workaround code to the openconceptlab module to ensure when it is saving new concepts from OCL that it preserves concept names as they are.  There was a similar issue there where it was adding concept names as they were listed in the import json, and that was resulting in the Concept addName API method changing the name in many cases (eg. if it's the first name added, make it an FSN, also sometimes making names preferred when they should not be, etc)

ibacher: Thanks @mseaton! Yeah, I think we need to work out a mechanism here to try and ensure names are added in the appropriate order. Of course, it'd be better if `addName()` didn't try to enforce these kinds of business rules itself.

Issue #177, opened 2022-02-14
'Appointmentservicetypes' domain should not fetch by name


reagan-meant:
'Appointmentservicetypes' domain should not fetch by name because name is not unique.

ibacher:
Fixed via #178.

PR #178, opened 2022-02-14
ISSUE-177: 'Appointmentservicetypes' domain should not fetch by name

reagan-meant:


ibacher (commented):
Thanks @reagan-meant! Code looks good. Two small things:

1. The issue is #177 not #170.
2. Does this work if no UUID is assigned? If not, we should update the docs to note that UUID needs to be assigned an throw an appropriate exception if none is provided.

reagan-meant:
@ibacher I think it is not a must to set a uuid

rbuisson:
> @ibacher I think it is not a must to set a uuid

@reagan-meant If that's not a must, then isn't it suppose to try fetch by name when no UUID is provided?
That does not make sense. We prob should make it mandatory to provide a UUID if we can't assume the name to be unique.

api/src/main/java/org/openmrs/module/initializer/api/appt/servicetypes/AppointmentServiceTypesCsvParser.java
```diff
@@ -34,18 +34,14 @@ public Domain getDomain() {
 	public AppointmentServiceType bootstrap(CsvLine line) throws IllegalArgumentException {
 		
 		final String uuid = line.getUuid();
-		
-		AppointmentServiceType type = StringUtils.isEmpty(uuid) ? null : service.getAppointmentServiceTypeByUuid(uuid);
-		
-		if (type == null) {
-			type = Utils.fetchBahmniAppointmentServiceType(line.getName(), service);
+		if (StringUtils.isEmpty(uuid)) {
+			throw new IllegalArgumentException("Uuid should be set");
```
rbuisson, line 38: Please use the same message format as here: https://github.com/mekomsolutions/openmrs-module-initializer/blob/7a2b7c6d532d3dc27457604c1a5308f0fcc311a4/api/src/main/java/org/openmrs/module/initializer/api/providerroles/ProviderRolesCsvParser.java#L33

rbuisson (changes requested):

api/src/test/resources/testAppDataDir/configuration/appointmentservicetypes/servicetypes.csv
```diff
@@ -1,5 +1,6 @@
 Void/Retire,UUID,Name,Service definition,Duration,_order:1000
-,,Short follow-up,On-site Appointment,10,
+,f378bec4-2d0d-4509-a56e-b709e0a53700,Short follow-up,On-site Appointment,10,
 TRUE,4e0f61df-d1f7-4cff-8d69-6264666daf3b,,,,
 ,ffd7e0f4-33f1-4802-b87a-8d610ba1132d,Simple Bracing,Specialized Appointment,45,
-,,Complex Bracing,6b220700-4ba2-4846-86a7-a2afa5b6f2eb,75,
+,0a4624a6-ca81-42e2-a1ab-8f8dca033b83,Short follow-up,Specialized Appointment,45,
+,2526359e-cb9e-45d4-b83a-2b588744eb05,Complex Bracing,6b220700-4ba2-4846-86a7-a2afa5b6f2eb,75,
```
ibacher, line 6:
```suggestion
,2526359e-cb9e-45d4-b83a-2b588744eb05,Complex Bracing,6b220700-4ba2-4846-86a7-a2afa5b6f2eb,75,

```

ibacher (approved):

Issue #101, opened 2021-04-29
Domain for Liquibase changesets


mks-d:

mseaton:
@reagan-meant - at PIH, in our PIH EMR distribution, we have some custom code to execute liquibase changsets that we run immediately before loading metadata from Initializer.  So I would expect that we would add a new "liquibase" domain to Initializer and that it would be the first domain loaded.

Since liquibase xml files can contain references to other liquibase xml files (using the include tag), I would not expect us to support loading any file in this directory, but rather support one specific file that the loader would load (called "liquibase.xml").  It would then be up to this loader to include other files at relative paths as appropriate.

So the simple implementation of this loader would extend BaseFileLoader, and would implement the load(File) method something like this:

```
if (the file being loads is named "liquibase.xml" and is at the root of the domain directory) {
  DatabaseUpdater.executeChangelog(file.getAbsolutePath(), null);
}
```

rbuisson:
Thanks for detailing your requirement and techincal approach @mseaton !

So to summarize:
- Add a `liquibase` domain
- Make sure it runs before every other domain
- Assume only 1 file in the domain: **liquibase.xml**. Loading additional files should be handled directly by the Liquibase engine.
- Add unit tests
- Add documentation.

rbuisson:
Hi @Ruhanga , I'm assigning you to this ticket. Priority is end of the month.
Is this ok with you?

mseaton:
Yes, that sounds good.  To be clear there may be other files in the domain directory, but the only file that the loader needs to directly concern itself with is the one called "liquibase.xml".

Ruhanga:
>Priority is end of the month.
Is this ok with you?

Fair enough @rbuisson. Thanks all.

PR #179, opened 2022-02-22
[#101]: Adding domain for Liquibase changesets

Ruhanga:
See #101

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
@@ -1,0 +1,72 @@
+package org.openmrs.module.initializer.api.loaders;
+
+import java.io.File;
+import java.sql.Connection;
+
+import org.openmrs.api.context.Context;
+import org.openmrs.module.initializer.Domain;
+import org.openmrs.module.initializer.api.ConfigDirUtil;
+import org.openmrs.util.DatabaseUpdater;
+import org.springframework.stereotype.Component;
+
+import liquibase.Liquibase;
+import liquibase.database.Database;
+import liquibase.database.DatabaseFactory;
+import liquibase.database.jvm.JdbcConnection;
+import liquibase.resource.FileSystemResourceAccessor;
+
+@Component
+public class LiquibaseLoader extends BaseFileLoader {
+	
+	public static final String LIQUIBASE_FILE_NAME = "liquibase";
+	
+	public static final String DATABASE_CHANGELOG_TABLENAME = "liquibasechangelog";
+	
+	public static final String DATABASE_CHANGELOGLOCK_TABLENAME = "liquibasechangeloglock";
+	
+	@Override
+	protected Domain getDomain() {
+		return Domain.LIQUIBASE;
+	}
+	
+	@Override
+	protected String getFileExtension() {
+		return "xml";
+	}
+	
+	@Override
+	protected void load(File file) throws Exception {
+		if (file.getName().equalsIgnoreCase(LIQUIBASE_FILE_NAME + "." + getFileExtension())) {
+			updateDatabase(file.getPath());
+			Context.flushSession();
+			Context.clearSession();
+		}
+	}
+	
+	@Override
+	public ConfigDirUtil getDirUtil() {
+		return new ConfigDirUtil(iniz.getConfigDirPath(), iniz.getChecksumsDirPath(), getDomainName(), true);
+	}
+	
+	private void updateDatabase(String filename) throws Exception {
+		Liquibase liquibase = getLiquibase(filename);
+		liquibase.update("");
+		liquibase.getDatabase().getConnection().commit();
+	}
+	
+	private Liquibase getLiquibase(String filename) throws Exception {
+		Connection connection = DatabaseUpdater.getConnection();
+		Database database = DatabaseFactory.getInstance().findCorrectDatabaseImplementation(new JdbcConnection(connection));
+		database.setDatabaseChangeLogTableName(DATABASE_CHANGELOG_TABLENAME);
+		database.setDatabaseChangeLogLockTableName(DATABASE_CHANGELOGLOCK_TABLENAME);
+		if (connection.getMetaData().getDatabaseProductName().contains("HSQL Database Engine")
+		        || connection.getMetaData().getDatabaseProductName().contains("H2")) {
+			// a hack because hsqldb and h2 seem to be checking table names in the metadata section case sensitively
+			// see https://github.com/openmrs/openmrs-core/blob/e679877103a89937a867d1fdf3aa48f8caaddd73/api/src/main/java/org/openmrs/util/DatabaseUpdater.java#L412-L417
+			database.setDatabaseChangeLogTableName(DATABASE_CHANGELOG_TABLENAME.toUpperCase());
+			database.setDatabaseChangeLogLockTableName(DATABASE_CHANGELOGLOCK_TABLENAME.toUpperCase());
+		}
+		
+		return new Liquibase(filename, new FileSystemResourceAccessor(), database);
+	}
+}
--- a/api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
@@ -1,28 +1,16 @@
 package org.openmrs.module.initializer.api.loaders;
 
 import java.io.File;
-import java.sql.Connection;
-
 import org.openmrs.api.context.Context;
 import org.openmrs.module.initializer.Domain;
 import org.openmrs.module.initializer.api.ConfigDirUtil;
 import org.openmrs.util.DatabaseUpdater;
 import org.springframework.stereotype.Component;
 
-import liquibase.Liquibase;
-import liquibase.database.Database;
-import liquibase.database.DatabaseFactory;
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
```diff
@@ -0,0 +1,72 @@
...
+	
+	@Override
+	protected String getFileExtension() {
+		return "xml";
+	}
+	
+	@Override
+	protected void load(File file) throws Exception {
+		if (file.getName().equalsIgnoreCase(LIQUIBASE_FILE_NAME + "." + getFileExtension())) {
+			updateDatabase(file.getPath());
```
mseaton, line 40: The only thing you should need to do here is defer to OpenMRS.  Simply use the "org.openmrs.util.DatabaseUpdater" class from OpenMRS core:
```
DatabaseUpdater.executeChangelog(file.getPath(), null);
```

mseaton (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
Ruhanga: Ohhh, yes indeed @mseaton. I missed this one, thank you very much for pointing it out.

readme/liquibase.md
```diff
@@ -0,0 +1,33 @@
+## Domain 'liquibase'
+The **liquibase** subfolder contains liquibase configuration files for the purpose of updating difficult to predetermine metadata before they can be consumed safely. The underlying files may contain or reference other files defining database changesets. This is how this domain may look like:
```
rbuisson, line 2: Let's make sure Liquibase is written with a capital L where appripriate (anything else than the folder or domain name).
And:
> contains liquibase configuration files

"contains Liquibase changelog files"

rbuisson (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
```diff
@@ -0,0 +1,38 @@
...
+	protected String getFileExtension() {
+		return "xml";
+	}
+	
+	@Override
+	protected void load(File file) throws Exception {
+		if (file.getName().equalsIgnoreCase(LIQUIBASE_FILE_NAME + "." + getFileExtension())) {
+			DatabaseUpdater.executeChangelog(file.getPath(), null);
+			Context.flushSession();
+			Context.clearSession();
```
mseaton, line 30: Did you determine that these calls to flush and clear session are necessary / useful?

readme/liquibase.md
```diff
@@ -0,0 +1,33 @@
+## Domain 'liquibase'
+The **liquibase** subfolder contains Liquibase changelog files for the purpose of updating difficult to predetermine metadata before they can be consumed safely. The underlying changelog files may contain or reference other files defining database changesets. This is how this domain may look like:
+
+```bash
+liquibase/
+  └── liquibase.xml
+  └── concepts.xml
+```
+The purpose for this domain is to provide updates on metadata and not any database schema. This for example, could be used to ensure all desired metadata is deterministic in nature with known `id`s/`uuid`s before it's/they are usable/reference-able.
```
mseaton, line 9: Why are we making this qualification?  We use liquibase in our configuration at PIH for all sorts of things - data migrations, creation of indexes, views, and functions for use in reports, etc.  To me the purpose is to allow implementations to execute database updates without having to write and deploy a heavy module just to deliver a liquibase changeset.  Let's leave the how and why to implementations to figure out.

readme/liquibase.md
```diff
@@ -0,0 +1,33 @@
+## Domain 'liquibase'
+The **liquibase** subfolder contains Liquibase changelog files for the purpose of updating difficult to predetermine metadata before they can be consumed safely. The underlying changelog files may contain or reference other files defining database changesets. This is how this domain may look like:
+
+```bash
+liquibase/
+  └── liquibase.xml
+  └── concepts.xml
```
mseaton, line 7: It needs to be much clearer that "liquibase.xml" is what is required and expected.  Something like "concepts.xml" that would be referred to is supported, sure, but it's not the primary use case and should not be the focus of the documentation.  At PIH, we only use 1 xml file, but it refers to many SQL files by relative path.  Might be better off just linking to the liquibase documentation with a blurb about how liquibase.xml is required as the starting point, but that liquibase allows for changesets to refer to other changesets and external resources.

mseaton (approved):
Looks good to me, though I did add a few comments for consideration.

api/src/test/java/org/openmrs/module/initializer/api/LiquibaseLoaderIntegrationTest.java
```diff
@@ -0,0 +1,44 @@
...
+	public void setup() {
+		System.setProperty("useInMemoryDatabase", "true");
+	}
+	
+	@Test
+	public void load_shouldLoadStructuredLiquibaseChangesets() throws Exception {
+		// Replay
+		loader.load();
+		
+		// Verif
```
ibacher, line 37:
```suggestion
		// Verify
```

api/src/test/resources/testAppDataDir/configuration/liquibase/liquibase.xml
```diff
@@ -0,0 +1,38 @@
...
+		<comment>
+			Set encounter type 'Laboratory' UUID
+		</comment>
+		<update tableName="encounter_type">
+			<column name="uuid" value="4bf982f0-8053-4757-a45e-5da777ffe0f6" />
+			<where>name='Laboratory'</where>
+		</update>
+	</changeSet>
+
+</databaseChangeLog>
```
ibacher, line 38:
```suggestion
</databaseChangeLog>

```

readme/liquibase.md
```diff
@@ -0,0 +1,33 @@
...
+
+###### liquibase.xml configuration file example:
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                  http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+	<include file="concepts.xml" relativeToChangelogFile="true"/>
+
+	<!-- Encounter Types -->
+	<changeSet id="jUnit-2022-02-22-12-51" author="test">
```
ibacher, line 19: At least in documentation, it's better to use a better example:

```suggestion
	<changeSet id="ensure_admission_encountertype_uuid-202202221251" author="test">
```

ibacher (changes requested):
Definitely need to look into why the [test for this is failing|https://github.com/mekomsolutions/openmrs-module-initializer/runs/5318128314?check_suite_focus=true#step:4:7442] on CI before this can be merged. Otherwise, just some minor comments.

readme/liquibase.md
Ruhanga: >This for example, could be used to ensure all desired metadata is deterministic in nature with known `id`s/`uuid`s before it's/they are usable/reference-able.

The above little doc is a simple example use-case that could help a non-liquibase acquainted  user of the module.

readme/liquibase.md
Ruhanga: A link to the Liuqibase docs is actually way much handy indeed. I'll add a reference link. Thanks @mseaton.

api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
Ruhanga: Thanks for the review @mseaton. I thought it safe since I've not gone through all edge-case scenarios. I'd be happy to remove those methods if this has not posed any problems based on your usage of `DatabaseUpdater .executeChangelog()`.

readme/liquibase.md
Ruhanga: Implementations are not restricted either from building their own complex Liquibase scripts in the end.

api/src/test/java/org/openmrs/module/initializer/api/LiquibaseLoaderIntegrationTest.java
Ruhanga: >[test for this is failing](https://github.com/mekomsolutions/openmrs-module-initializer/runs/5318128314?check_suite_focus=true#step:4:7442)

Tests pass locally and there are no related runtime excepts. I'm not sure why this is only happening with github-actions. Would you help look into this @rbuisson, maybe @ibacher? Thanks.

api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
mseaton: @Ruhanga yes, I would say remove those lines.  We don't do anything like that.  Liquibase doesn't have anything to do with Hibernate - it executes directly against a separate database connection - so flushing and clearing the Hibernate session is not something that should be done here.

ibacher (approved):

mks-d:
@ibacher / @Ruhanga this will require a follow-up commit that provides all the necessary README additions and creations.

ibacher:
@mks-d Whoops. Done in 1a8dcd4.

Issue #157, opened 2021-11-06
Allow INFO, DEBUG, and TRACE level logging to occur


mseaton:
I may be missing something obvious, but there is nothing I can do to get any of my INFO, DEBUG, and TRACE level logging to appear in the Console when I am developing in Iniz.  I see there are some aspects of the codebase that are hard-coding WARN levels, which I suspect is the issue, though I haven't been able to successfully undo that.

In general, we should not be hard-coding any particular logging.  If we want to support a separate initializer.log file that focuses only on the WARN messages from Iniz, this should be something that is possible without affecting other appenders and logging mechanisms.

@mks-d thoughts?  This may be related to #67 though not 100% clear.

mks-d:
@mseaton you meant this? https://github.com/mekomsolutions/openmrs-module-initializer/blob/a10d37e6990827edf37a60a4c51c3fe76d3978f9/api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java#L47

And yes, I agree that having all this configurable externally would definitely be better.

mseaton:
@mks-d yes, that's what I'm referring to.  The Validator.java class also has a "setupLog4j" method that manipulates things to hard-code to WARN messages as well.  Some of this is likely OK, but what I've found is that I can't get any logging to appear in the console unless I use WARN levels, which is hard for development, as it means I either avoid logging or I have to do big copy/replace operations before I commit.

mseaton:
This seems like it might be getting addressed in PR #215 ?  @mks-d and @icrc-fdeniger , should we link that PR to this ticket?  Will that address the concerns laid out above?

mks-d:
>For your issue, I would expect to have a dedicated log4j config file that will be used in dev mode

@icrc-fdeniger do you mean that there is a workaround and that this issue could be closed?

icrc-fdeniger:
IMHO this issue is still here. This line: 
https://github.com/mekomsolutions/openmrs-module-initializer/blob/a10d37e6990827edf37a60a4c51c3fe76d3978f9/api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java#L47
will prevent displaying INFO, DEBUG, and TRACE level logging as mentionned by mseaton.

ibacher:
[...]

PR #181, opened 2022-02-28
Support for 2.4.0 / Log4J2

### LINKED ISSUE #175: Support for 2.4.0+ / Log4J2
Currently, Iniz hard-codes setting up a logger in the Activator [here](https://github.com/mekomsolutions/openmrs-module-initializer/blob/5b2c7a24879cc24bc8f86c49eb731b6242f340cc/api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java#L46).

This will fail on versions of OpenMRS that use Log4J2 (2.4.0+ and now 2.1.5+) because the class `org.apache.log4j.FileAppender` is no longer on the classpath. (Log4J2 provides a Log4J1 "bridge" that maps many Log4J1 class, but reimplements them on Log4J2's infrastructure).

We should refactor things so it's possible to load Iniz on versions of OpenMRS using Log4J2 while maintaining the current functionality on versions that use Log4J1.

ibacher:
Fixes #175. This actually took less time than I thought as I realised that the Iniz needs are much simpler. This should also address #157.

To do:
- [x] Testing
- [x] Documentation


ibacher:
With the caveat that I'm still fighting to get the Validator to run on my machine, I think this has all the necessary changes. Certainly the module itself runs correctly.

gracepotma:
@rbuisson or @ibacher could you assign a reviewer? Sorry if I'm poking my nose in unnecessarily; anything that has "Log4j2" makes me uncomfortable left open :P

ibacher:
@gracepotma Nothing to do with the Log4J2 vulnerability... Initializer just internally does some configuration to generate it's own log file.

Issue #145, opened 2021-10-15
Support ordered loading of message.properties files


mogoodrich:
Like the "_order" property that can be used to specify the order in which csv-based metadata is loaded, we'd like to be able to order the way messages.properties files are loaded by Iniz.

Use case:  In the PIH EMR, country-specific implementations want to override a limited number of messages properties... ie the Peru implementation wants a slightly different Spanish translation for Consult Note than our Mexico implementation.  We'd like to ship a standard "messages_encountertypes_es.properties" with all our implementations, and then ship a specific "messages_encountertypes_peru_es.properties" that overrides the name of this single encounter type.  I think we can do by specifying the order in which messages properties are loaded, and then enforcing this loading order here.

We could support this by allowing the first field in a messages.properties file to be "_order", a la:

```
_order:1000
my.code: My Translation
my.other.code: My Other Translation

```

brandones:
@mks-d What do you think of this? I think we (PIH) could implement it if it seems good to you.

mks-d:
It's like CSV files right, there is no proper way to attach metadata to Java properties file, so we need to hack it into its content.

Will anybody ever need the key `_order`? I have to admit that most likely not.

I guess my issue is more that we generate an actual key-value pair. In the case of CSVs we don't interfere with the values under the header, so that if anyone ever needed that header to provide values, they could. The Java properties equivalent would be more something like:
```
_order_1000:
```
And if anybody ever needed that improbable key, well they still could.

I am certainly overthinking this, feel free to implement the suggested solution.

mseaton:
@mogoodrich and @brandones - so we still need this feature?  With the new message source implemented in #155 I would think that, at least for our needs at PIH, we can achieve this easily by just having messages_es.properties that can be specified further via messages_es_PE.properties and messages_es_MX.properties, and then ensuring that the default and allowed locales used by Peru and Mexico are es_PE and es_MX respectively.  This wouldn't have worked reliably before some of the revisions to the message source in #155 but should do so if/when that is merged.

brandones:
We don't, since we only run one implementation per country. We're sort of abusing the country coded files, because the things in "es_PE" aren't actually differences between Peruvian and Mexican Spanish, but idiosyncrasies of the Peru implementation of OpenMRS. I think having an organic way to define message code "overrides" would be valuable.

mogoodrich:
[...]

PR #183, opened 2022-03-21
[#145] Add support for messageproperties to support ordered loading through special _order message code

mseaton:


Ruhanga (approved):
Looks good @mseaton, thanks for this intuitive solution. Where as the motivation might have been that of _"overriding properties"_ , this might also come in handy for _"message property interpolation"_.

readme/messageproperties.md
```diff
@@ -9,5 +9,11 @@ messageproperties/
 There can be as many message properties internationalization files as needed.
 This domain differs from most others since nothing from its configuration is persisted in database, everything is stored in the runtime memory upon starting the Initializer.
 
+If one needs to configure the order in which message properties files load, for example if one message properties
+file contains message codes that should override those in another in the same locale, one can do this by utilizing a 
+special message code named "_order".  Two message properties files in the same locale will be loaded in order based
+first on the value of any "_order" property defined, and second based on the alphabetical order of the 
+absolute path of the file.
```
mks-d, line 16: @mseaton could you document what happens if `_order` is specified in only one of two files for the same locale?

I suspect that one will be loaded first?

mks-d (commented):
I would kindly request a commit headline à la
>[#145] messageproperties to support ordering through special _order property key.

... or whatever you deem appropriate.

readme/messageproperties.md
mseaton: Done.  Let me know if that's clear.

PR #192, opened 2022-07-27
Correct documentation about concept FSNs

brandones:
Concepts do not need one FSN per locale; they only need one FSN. It does not even have to be in the default locale.


brandones:
Stale; closing.

Issue #193, opened 2022-07-29
Concepts domain should support other mapping types


brandones:
Right now, the concepts domain only supports SAME-AS mappings. It should support others, such as NARROWER-THAN.

I propose that the headers should be changed to `Mappings|SAME-AS`, `Mappings|NARROWER-THAN`, etc. `Same as mappings` would be interpreted as `Mappings|SAME-AS` for backwards compatibility.

mogoodrich:
+1 thanks @brandones

mseaton:
Instead of cluttering up the headers, which could make for some very noisy and hard to read csvs, what about keeping the existing concept csv mapping headers to assume SAME-AS, and recommend use for things we consider codes and consistent across all concepts (like the PIH and CIEL mappings).  Then, we'd have a new domain - conceptreferenceterm (or similar), which would have a CSV with columns:

concept | source | term | type

This would allow multiple csvs (one per source) or a single CSV that contains multiple source/mapping pairs.

@mks-d / @rbuisson thoughts?

mks-d:
@mseaton very much like with concept names I believe the two options can certainly coexist. Especially if implementers see value in @brandones's suggestion.
Your proposal is definitely better for being more complete at properly addressing the details of those mapping types in the data model.

A bit of a non-answer I know, which means that we're supportive of both options and leave it up to you to implement first (and perhaps only) the one that you prefer.

mseaton:
That makes sense to me.  I'll create a new ticket around the conceptreferenceterm domain as needed.  Leaving this ticket as enhancing the concept domain, I'd like to modify the existing proposal on this ticket to support the following as supported header formats:

MAPPING-TYPE

* If the header just contains the mapping type (SAME-AS, NARROWER-THAN, etc), then this would lead to the exact same behavior as now (the existing "Same as mappings" would be mapped to "SAME-AS").  This is basically the original proposal in the ticket, with the exception that Mappings| is removed from the column header, as it can be implied by the inclusion of the mapping type.

MAPPING-TYPE|SOURCE"

* Optionally, the header can contain the specific source.  So you might have headers for "SAME-AS|CIEL", "SAME-AS|PIH", and "SAME-AS|RxNorm".  In this case, each of these would be expected to contain a single mapping.

MAPPING-TYPE|SOURCE|UNIQUE-KEY

[...]

PR #196, opened 2022-08-04
[#193] Expanded concept mappings capabilities.

mseaton:
Expand concepts domain to support concept mappings other than same-as, and to enable mappings to be specified in multiple columns if desired.

See #193 for details.


api/src/main/java/org/openmrs/module/initializer/api/c/MappingsConceptLineProcessor.java
```diff
@@ -4,36 +4,75 @@
...
 import org.springframework.stereotype.Component;
 
 @Component("initializer.mappingsConceptLineProcessor")
 public class MappingsConceptLineProcessor extends ConceptLineProcessor {
 	
-	protected static String HEADER_MAPPINGS_SAMEAS = "same as mappings";
+	protected final Logger log = LoggerFactory.getLogger(getClass());
 	
-	private ConceptMapListParser listParser;
+	public static final String MAPPING_HEADER_PREFIX = "mapping";
```
mks-d, line 23: I would favour here to go for plural since those headers accept arrays of values:
```java
public static final String MAPPINGS_HEADER_PREFIX = "mappings";
```
and so one for associated headers.

api/src/main/java/org/openmrs/module/initializer/api/c/MappingsConceptLineProcessor.java
```diff
@@ -4,36 +4,75 @@
...
+			
+			String lineValue = line.get(header);
+			if (!StringUtils.isEmpty(lineValue)) {
+				
+				// For backwards-compatibility, support the original same as mappings column header
+				if (header.trim().equalsIgnoreCase(HEADER_MAPPINGS_SAMEAS)) {
+					header = MAPPING_HEADER_PREFIX + MAPPING_HEADER_SEPARATOR + ConceptMapType.SAME_AS_MAP_TYPE_UUID;
+				}
+				
+				String[] headerComponents = header.split(MAPPING_HEADER_SEPARATOR_REGEX);
```
ibacher, line 52: I think this is slightly preferable...

```suggestion
				String[] headerComponents = header.split(MAPPING_HEADER_SEPARATOR_REGEX, 3);
```

api/src/main/java/org/openmrs/module/initializer/api/c/MappingsConceptLineProcessor.java
```diff
@@ -4,36 +4,75 @@
...
+				
+				if (headerComponents[0].equalsIgnoreCase(MAPPING_HEADER_PREFIX)) {
+					
+					// First determine the map type (required)
+					String headerMapTypeStr = headerComponents[1].trim().toLowerCase();
```
ibacher, line 57: So, OMRS can be configured to run in either case-sensitive or case-insensitive mode for database searches; this seems to require OMRS be running in case-insensitive mode. Do we need this normalisation to lower case?

ibacher (approved):
Looks great @mseaton! Two small things I might adjust, but I don't see any major issues with it.

brandones:
Updating the exporter: https://github.com/PIH/iniz-exporters/pull/4

api/src/main/java/org/openmrs/module/initializer/api/c/MappingsConceptLineProcessor.java
mseaton: I had this originally, but changed it to be less rigid.  There are actually up to 4 components (mappings|type|source|unique-key), though only 2 components are required.

api/src/main/java/org/openmrs/module/initializer/api/c/MappingsConceptLineProcessor.java
mseaton: That's a good point.  I tend to take case-insensitive db searching as a given in openmrs.  We need the header to be case-insensitive I think to be consistent with every other header in iniz.  But defer to @mks-d .  I think this is likely an issue throughout the entire module, not just with this particular component.

api/src/main/java/org/openmrs/module/initializer/api/c/MappingsConceptLineProcessor.java
mseaton: @mks-d I have pushed a new commit that makes this change to "mappings" plural

api/src/main/java/org/openmrs/module/initializer/api/c/MappingsConceptLineProcessor.java
mseaton: @ibacher I have pushed a new commit that limits the split to 4 components.

mseaton:
I think I have addressed the PR comments.  Let me know if I'm good to merge this in or if this needs more review or one of you would prefer to merge it in when ready.  Thanks @mks-d and @ibacher

Issue #198, opened 2022-09-07
Reloading Concept CSVs should not clear Member and Answers if those columns aren't in the CSV


mogoodrich:
We added support for concept sets domain here:

https://github.com/mekomsolutions/openmrs-module-initializer/issues/140

We should now implement the functionality suggested in a comment at the end of that ticket:

"One thought I had was that, if a "concepts.csv" domain file has an "Answers" or "Members" column defined, and if that column is empty for a given Concept row, then clear existing Answers/Members as appropriate. If the CSV does not have this column, do nothing. But if the CSV has the column and a Concept leaves values blank, then infer that this means it should have no members/answers. Thoughts on this modification?"

Currently, I believe that whenever a concept file is reloaded, all the members/answers are reloaded *regardless* of whether or not the Member or Answers column is present (note: I've tested this behavior for members, but not for answers).

This is incredibly dangerous, because when using the concepts.csv + concept_sets.csv module, then any time a change is made to a concepts.csv file, *all* concepts are removed from any sets or answered in that file unless a corresponding change happens to be made to the set files for all the concepts in that csv.

fyi @mseaton

mseaton:
PR merged, closing this out.

PR #199, opened 2022-09-23
[#198] Reloading Concept CSVs should not clear Member and Answers if …

mogoodrich:
…those columns aren't in the CSV

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/c/NestedConceptLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/c/NestedConceptLineProcessor.java
@@ -29,7 +29,7 @@ public class NestedConceptLineProcessor extends ConceptLineProcessor {
 	
 	public Concept fill(Concept concept, CsvLine line) throws IllegalArgumentException {
 		
-		if (!CollectionUtils.isEmpty(concept.getAnswers())) {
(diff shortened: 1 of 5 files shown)


mseaton (approved):
This looks good to me - regardless of what other improvements we might make in the future, I think this is the right thing to do.  However, as things are starting to get rather complex with various types of metadata affected my multiple iniz domains, I think more discussion is needed overall as to how we approach this.  We can discuss outside of this ticket.

mogoodrich:
Just pinging you on this one to take a look when you get a chance @ibacher ... thanks!

Ruhanga (approved):

mogoodrich:
Just another ping @ibacher ... :)  fyi @mseaton if you want

mseaton:
@mogoodrich I think we are likely good to merge this, but can you push an update to the concepts.md file in the readme that documents the behavior for the members and answers?  eg. describe that - if the columns are present, this will clear any existing members and answers and add those specified.  If the columns are not present, it will not change any existing relationships?

mogoodrich:
Fair enough @mseaton , will do.

ibacher (commented):
LGTM

ibacher (approved):

mogoodrich:
@mseaton documentation updated

PR #200, opened 2022-09-27
Fix Liquibase Loader Test

mogoodrich:


mogoodrich:
I noticed when running tests that the LiquibaseLoaderIntegrationTest was failing for me when running all the tests via Maven.  Looking at the test, I saw a comment and some code ignoring when running github actions.

I think I figured out the error: the DeleteDomainChecksumsChangesetIntegrationTest was polluting the context somehow, and so if it happened to run before the LiquibaseLoaderIntegrationTest the latter fails.

I was able to fix it by changing the name the liquibase files used (see PR) though the "proper" fix would likely be to fix the DeleteDomain test to clean up it's context.

fyi @ibacher

mogoodrich:
But now github actions is failing on this test:
"load_shouldLoadFormWithAllAttributesSpecified(org.openmrs.module.initializer.api.form.HtmlFormsLoaderIntegrationTest)"
... but I'm not seeing this locally... so this may not help things...

mks-d (commented):
@mogoodrich is this still current?

PR #201, opened 2022-09-29
Add domain for FHIR ObservationCategoryMaps

ibacher:


api/src/main/java/org/openmrs/module/initializer/api/fhir/ocm/FhirObservationCategoryMapLineProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+			throw new IllegalArgumentException(
+			        "'" + CONCEPT_CLASS_HEADER + "' was not found for FHIR observation category map " + instance.getUuid());
+		}
+		
+		ConceptClass cc = conceptService.getConceptClassByName(conceptClass);
+		if (cc == null && !instance.getRetired()) {
+			throw new IllegalArgumentException("Concept class " + conceptClass
+			        + " was not found while creating FHIR observation category map " + instance.getUuid());
+		}
+
```
Ruhanga, line 54: Thanks @ibacher for this PR. Wouldn't it be fairer fetching the ConceptClass by uuid(preferred) and alternatively as a fall back, fetching it by name? This is because the name could change especially for custom concept classes for some non-trivial reason. I'd lean towards fetching an entity by all it's unique identifiers before it can be marked _'not found'_.

ibacher: @Ruhanga Yeah, that's reasonable. I suppose the way I think about Iniz is that it should be making the metadata _readable_ first (i.e., by looking at a file, I should be able to work out what it will do without needing to know too much about OpenMRS internals), but i maybe makes sense to support UUIDs too. (So UUIDs can be supported, but maybe not _preferred_).

api/src/main/java/org/openmrs/module/initializer/api/fhir/ocm/FhirObservationCategoryMapCsvParser.java
```diff
@@ -0,0 +1,75 @@
...
+			                + ConceptClass.class.getSimpleName() + " cc where cc.name = :conceptClassName" + ")")
+			        .setParameter("fhirObsCategory", fhirObsCategory).setParameter("conceptClassName", conceptClass)
+			        .uniqueResult();
+		}
+		
+		if (result == null && line.getUuid() != null && !line.getUuid().isEmpty()) {
+			result = (FhirObservationCategoryMap) sessionFactory.getCurrentSession()
+			        .createQuery("from " + FhirObservationCategoryMap.class.getSimpleName() + " where uuid = :uuid")
+			        .setParameter("uuid", line.getUuid()).uniqueResult();
+		}
```
Ruhanga, line 60: @ibacher, I'd have this logic decoupled from iniz, so having it in the fhir2 dao/service would be fairly neater if there is no limitation for this.

api/src/main/java/org/openmrs/module/initializer/api/fhir/ocm/FhirObservationCategoryMapLineProcessor.java
Ruhanga: This is fine with me. Thannks.

mks-d (commented):
@mherman22 could you provide a detailed description to this PR so as to provide more context? Or @ibacher could you speak to it?

ibacher:
So, FHIR Observations have a `category` property that tries to cover "what kind" of observation that they are. In OMRS, the same concept is usually represented by "concept class" for the concept associated with the observation. This provides a way of mapping between the FHIR `category` field and the concept's concept class because, wherever possible, we want to use the [standard FHIR value set](http://hl7.org/fhir/R4/valueset-observation-category.html), but OMRS's ConceptClasses don't necessarily always align with those categories.

Basically, we use this field to power the "lab results viewer" and could use it for other things, like vital signs, but we need a way of allowing custom mappings to accommodate this flexibly.

mks-d (approved):
LGTM, two things:
1. @ibacher could you address the conflict reported?
2. @Ruhanga could you confirm that your requested change has been addressed by validating this PR?

ibacher:
I still need to move the actual logic into the FHIR2 module, so this remains pending.

PR #214, opened 2022-12-21
Bump version to 2.5.0-SNAPSHOT

rbuisson:


mks-d (changes requested):
You need to move all 2.4.1 release notes (there's only one) to 2.5.0 since there was never a release 2.4.1.

rbuisson:
Makes sense. Pushed the fix.

mks-d (approved):

PR #216, opened 2023-02-02
Fix logging to use logger levels instead of filters

icrc-fdeniger:
PR to Fix an issue included in my previous PR concerning log management:
in core 2.3, log level is configured using reflection as  some classes are not available.
In the previous PR I used a wrong method signature for `org.apache.log4j.varia.LevelRangeFilter`.
This PR will fix this issued and add a Unit Test.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0.java
@@ -44,12 +44,7 @@ public class InitializerLogConfigurator2_0 implements InitializerLogConfigurator
 			        .newInstance(layout, logFilePath.toString());
 			appender.setName(logFilePath.getFileName().toString());
 			
-			// since the org.apache.log4j.varia package doesn't exist in the log4j2 bridge, we need to use reflection
-			// so this class only has a runtime dependency on the LevelMatchFilter
-			Filter levelRangeFilter = (Filter) Class.forName("org.apache.log4j.varia.LevelRangeFilter").getConstructor()
-			        .newInstance();
-			Method levelMatchSetter = levelRangeFilter.getClass().getMethod("setLevelMax", String.class);
-			levelMatchSetter.invoke(levelRangeFilter, level.toString());
+			Filter levelRangeFilter = createLevelRangeFilter(level);
 			
 			appender.addFilter(levelRangeFilter);
 		}
@@ -59,5 +54,16 @@ public class InitializerLogConfigurator2_0 implements InitializerLogConfigurator
 		}
 		
 		return appender;
+	}
+	
+	Filter createLevelRangeFilter(Level level) throws InstantiationException, IllegalAccessException,
+	        InvocationTargetException, NoSuchMethodException, ClassNotFoundException {
+		// since the org.apache.log4j.varia package doesn't exist in the log4j2 bridge, we need to use reflection
+		// so this class only has a runtime dependency on the LevelMatchFilter
+		Filter levelRangeFilter = (Filter) Class.forName("org.apache.log4j.varia.LevelRangeFilter").getConstructor()
+		        .newInstance();
+		Method levelMatchSetter = levelRangeFilter.getClass().getMethod("setLevelMax", Level.class);
+		levelMatchSetter.invoke(levelRangeFilter, level);
+		return levelRangeFilter;
 	}
 }
--- a/api/src/main/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0.java
@@ -30,10 +30,11 @@ public class InitializerLogConfigurator2_0 implements InitializerLogConfigurator
 		
 		org.apache.log4j.Logger logger = org.apache.log4j.Logger
 		        .getLogger(InitializerActivator.class.getPackage().getName());
-		logger.addAppender(getFileAppender(level, logFilePath));
+		logger.addAppender(getFileAppender(logFilePath));
+		logger.setLevel(level);
 	}
 	
-	private Appender getFileAppender(Level level, Path logFilePath) {
+	private Appender getFileAppender(Path logFilePath) {
 		Appender defaultAppender = org.apache.log4j.Logger.getRootLogger().getAppender("DEBUGGING_FILE_APPENDER");
 		Layout layout = defaultAppender == null ? new PatternLayout("%p - %C{1}.%M(%L) |%d{ISO8601}| %m%n")
 		        : defaultAppender.getLayout();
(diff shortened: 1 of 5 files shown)


api/src/test/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0Test.java
```diff
@@ -0,0 +1,25 @@
...
+	public void createLevelRangeFilter_shouldCreateLevelWithMaxValue() throws ClassNotFoundException, InvocationTargetException,
+	        InstantiationException, IllegalAccessException, NoSuchMethodException {
+		InitializerLogConfigurator2_0 logConfigurator20 = new InitializerLogConfigurator2_0();
+		Level level = Level.WARN;
+		
+		Filter levelRangeFilter = logConfigurator20.createLevelRangeFilter(level);
+		
+		Assert.assertEquals(level, ((LevelRangeFilter) levelRangeFilter).getLevelMax());
+		
+	}
```
mks-d, line 23: Maybe?
```java

@Test
public void createLevelRangeFilter_shouldCreateLevelWithMaxValue() throws ClassNotFoundException, InvocationTargetException,
        InstantiationException, IllegalAccessException, NoSuchMethodException {
    // setup
    InitializerLogConfigurator logConfigurator = new InitializerLogConfigurator2_0();

    // replay
    Filter levelRangeFilter = logConfigurator.createLevelRangeFilter(Level.WARN);

    // verif
    Assert.assertEquals(level, ((LevelRangeFilter) levelRangeFilter).getLevelMax());
}
```

api/src/main/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0.java
```diff
@@ -60,4 +55,15 @@ private Appender getFileAppender(Level level, Path logFilePath) {
 		
 		return appender;
 	}
+	
+	Filter createLevelRangeFilter(Level level) throws InstantiationException, IllegalAccessException,
+	        InvocationTargetException, NoSuchMethodException, ClassNotFoundException {
+		// since the org.apache.log4j.varia package doesn't exist in the log4j2 bridge, we need to use reflection
+		// so this class only has a runtime dependency on the LevelMatchFilter
+		Filter levelRangeFilter = (Filter) Class.forName("org.apache.log4j.varia.LevelRangeFilter").getConstructor()
+		        .newInstance();
```
mks-d, line 64: I will leave it to @ibacher to validate this bit.

mks-d (commented):
I provided some minor comments.

Would it be possible for the PR title to describe what the change made to the code base actually is.

api/src/test/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0Test.java
icrc-fdeniger: here you miss the variable _level_ no ?

icrc-fdeniger: @mks-d will you prefer this approach :) :
```java
@Test
	public void createLevelRangeFilter_shouldCreateLevelWithMaxValue() throws ClassNotFoundException, InvocationTargetException,
	        InstantiationException, IllegalAccessException, NoSuchMethodException {
		InitializerLogConfigurator2_0 logConfigurator20 = new InitializerLogConfigurator2_0();

		Filter levelRangeFilter = logConfigurator20.createLevelRangeFilter(Level.WARN);

		Assert.assertEquals(Level.WARN, ((LevelRangeFilter) levelRangeFilter).getLevelMax());

	}
```

mks-d: Is create `createLevelRangeFilter` specific to 2.0+? If yes then fine.

I have always tried to push (junior) developers to divide their unit tests following the _given/when/then_ (or _setup/replay/verify_) pattern to clearly provide the reader with a millisec way to understand what the test scenario is. Of course this one is obvious, but dropping the guard is the best way to invite them all, including the more junior ones, to drop their guard too.

mks-d: @icrc-fdeniger correct, that would be it:
```java
@Test
public void createLevelRangeFilter_shouldCreateLevelWithMaxValue() throws ClassNotFoundException, InvocationTargetException,
        InstantiationException, IllegalAccessException, NoSuchMethodException {
    // setup
    Level level = Level.WARN;
    InitializerLogConfigurator logConfigurator = new InitializerLogConfigurator2_0();

    // replay
    Filter levelRangeFilter = logConfigurator.createLevelRangeFilter(level);

    // verif
    Assert.assertEquals(level, ((LevelRangeFilter) levelRangeFilter).getLevelMax());
}
```

icrc-fdeniger: Yes `createLevelRangeFilter` is specifc to 2.0. In fact it's a Unit test for `InitializerLogConfigurator2_0` so I named it the way.

For the Unit Test: the comments are missing right ?

api/src/main/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0.java
```diff
@@ -60,4 +55,15 @@ private Appender getFileAppender(Level level, Path logFilePath) {
 		
 		return appender;
 	}
+	
+	Filter createLevelRangeFilter(Level level) throws InstantiationException, IllegalAccessException,
```
mks-d, line 59: Any reason for the package visibility, rather than private I mean?

mks-d (commented):
A last tiny comment about the package visbility.

Could you please set a PR title that describes the change made to the code base? As in 
> Added this or that to InitializerLogConfigurator2_0 in order to...

So that the commit history remains a story teller.

api/src/main/java/org/openmrs/module/initializer/api/logging/InitializerLogConfigurator2_0.java
icrc-fdeniger: To be able to test it

ibacher:
@icrc-fdeniger I've somewhat hijacked this PR, but hopefully for the best...

ibacher:
Alright, I've slightly hijacked this PR to fix what I think the actual issue is.

In #181, for reasons that are now obscure to me, I added the `LevelMatchFilter` to the file appender in place of what had been a simple call to `logger.setLevel()`. I really have no idea why I went that route. The `LevelMatchFilter` doesn't work the way I anticipated: it literally only allows one logging level through (I imagine the use-case for this is to be able to shunt, e.g., DEBUG level logs to a noisy, and frequently refreshed file). The `LevelRangeFilter` is, of course, better (though the logic should be that the level we set is the minimum allowed through), but the simplest thing to do is... just call `logger.setLevel()`.

So, this PR now switches things over to using `logger.setLevel()` and adds some tests to ensure that the level _on the logger_ is set correctly.

icrc-fdeniger:
thanks @ibacher :)

Issue #211, opened 2022-11-30
Concepts should be fetched by FSN only


mks-d:
####  1) Background

See
* [Initializer: Matching Concepts by Name and the OpenMRS getConceptByName method ](https://talk.openmrs.org/t/initializer-matching-concepts-by-name-and-the-openmrs-getconceptbyname-method/37413?u=mksd)

####  2) Dev Notes
This happens [here](https://github.com/mekomsolutions/openmrs-module-initializer/blob/17c0420214e8f2a72e3bda1be370da3eff6c1f4a/api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java#L180) in `Utils#fetchConcept`.

####  3) Proposed Solution
After exposing failing unit tests that demonstrate that `ConceptService#getConceptByName` produces unreliable results, proceed to introduce a new method/service layer/DAO layer hosted in Iniz that ensures that only FSNs are being searched. This could largely be created on the same premises of the existing code, see [here](https://github.com/openmrs/openmrs-core/blob/de56986f5a788c551666e41ac2f2d100a19d7041/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java#L1752-L1767) in `ConceptService`'s DAO layer.

mseaton:
@mks-d when we do this, let's get rid of the hacky code that iterates over all of the Locales and sets the Context locale, in order to be able to search by concept in that locale.  Let's just create a DAO/Service method that will search by FSN without consideration of Locale.

mks-d:
@ibacher / @Ruhanga – what about this one?

ibacher:
Looks to me like it's implemented.

PR #218, opened 2023-02-17
[#211] Concepts should be fetched by FSN only

Ruhanga:
Issue: https://github.com/mekomsolutions/openmrs-module-initializer/issues/211


mseaton (changes requested):
This isn't exactly what I had in mind @Ruhanga .  See the comment / request I made on the ticket here:  https://github.com/mekomsolutions/openmrs-module-initializer/issues/211#issuecomment-1332595549.

We should implement this service method to do - as efficiently and correctly as possible - that which we need it to do, which is to retrieve a concept uniquely by FSN without regard for Locale.  So we should just do that in the DAO directly.

Beyond this, it feels a bit limiting to me for us to create such a specific service for just one method.  Since the need for specific DAO methods and services is relatively rare, it would be nice to just have an InitializerService...Impl...DAO...HibernateDAO, that could contain an assortment of service methods and DAO implementations as needed by the various domains in the module, rather than have these limited within the Concept domain.  If, one day, we find we need to add an improved service method for some other domain, it would be nice not to have to write another 4 classes and wire them all up to the Context and such.  I'd rather just write a new method in the existing (generic) classes.

We can debate the last point (not sure if @mks-d or @rbuisson have different perspectives), but I'd like to improve the DAO method as I described in the ticket unless we made the decision not to do so deliberately.

Ruhanga (commented):
Thanks @mseaton for the review. I agree with the suggestions. I have one minor question, is it possible that two concepts in different locales at one time can have deferent meanings for the same literal name/word in which case we'd have to decide which concept is required? Could this be a real world scenario that we'd need to address here or how would you suggest we handle such a case?

mseaton:
@Ruhanga - I'd suggest in this case we just fail hard - throw an exception indicating that the given Concept row cannot be loaded because the concept name matches the FSN of more than one concept.  If a name is ambiguous in terms of which Concept it represents, then it should not be used/allowed as a valid identifier.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -47,6 +47,9 @@ public class InitializerServiceImpl extends BaseOpenmrsService implements Initia
 	
 	private Map<String, Object> keyValueCache = new HashMap<String, Object>();
 	
+	@Autowired
```
mseaton, line 50: See above, don't autowire this in, but wire it into the existing service bean definition in the moduleApplicationContext.xml file.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -205,4 +208,8 @@ public Boolean getBooleanFromKey(String key) {
 	public InitializerConfig getInitializerConfig() {
 		return cfg;
 	}
+	
+	public Concept getConceptByName(String name) {
```
mseaton, line 212: Javadoc

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -177,7 +178,7 @@ public static Concept fetchConcept(String id, ConceptService service) {
 		try {
```
mseaton, line 178: This entire try/finally block along with the line above should be removed and replaced with a call to "service.getConceptByName(id)" only.  There is no longer any need to iterate over Locales and to set/reset the context locale.  So all this should be removed.

api/src/test/java/org/openmrs/module/initializer/api/ConceptsLoaderIntegrationTest.java
```diff
@@ -496,6 +496,22 @@ public void load_shouldLoadConceptNamesAccordingToCsvFiles() {
 		assertName(yellow, "Y", localeEn, false, shortName);
 	}
 	
+	@Test
+	public void load_shouldLoadConceptSetMembersByFullySpecifiedNameSearch() {
+		// Setup
+		Context.setLocale(localeEn);
```
mseaton, line 502: Setting the context locale here should no longer be needed for this test.  Remove this to eliminate confusion.

api/src/test/java/org/openmrs/module/initializer/api/ConceptsLoaderIntegrationTest.java
```diff
@@ -496,6 +496,22 @@ public void load_shouldLoadConceptNamesAccordingToCsvFiles() {
...
+		// Replay
+		loader.load();
+		
+		// Verify
+		// Retrieving the 'Overall vital signs' set concept that should have the actual 'Vital signs' concept 
+		// and not the 'False vital signs' concept
+		Concept actualVitalSigns = cs.getConceptByUuid("23542fd3-4315-4e51-b68e-bce887331c0a").getSetMembers().get(0);
+		Assert.assertNotNull(actualVitalSigns);
+		assertName(actualVitalSigns, "Vital signs", localeEn, true, ConceptNameType.FULLY_SPECIFIED);
+	}
```
mseaton, line 513: Please make sure that the existing unit tests cover the different cases that your service now handles.  eg. You should make sure there are tests that things will fail if 2 concepts have the same FSN in different locales.  You should make sure that a row will be found by FSN regardless of locale (eg. if the name locale is different from the current Context locale).

mseaton (changes requested):
Hi @Ruhanga .  Looking better.  A few changes requested.  See comments.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,64 @@
...
+			criteria.add(Restrictions.ilike("name", name));
+		} else {
+			criteria.add(Restrictions.eq("name", name));
+		}
+		
+		criteria.add(Restrictions.eq("voided", false));
+		criteria.add(Restrictions.eq("conceptNameType", ConceptNameType.FULLY_SPECIFIED));
+		
+		criteria.createAlias("concept", "concept");
+		criteria.add(Restrictions.eq("concept.retired", false));
```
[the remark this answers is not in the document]
ibacher: @mseaton Semantically, I'm not sure it makes sense to unretire a concept by FSN. The FSN, AFAICT, is only guaranteed to be unique for unretired concepts, so for a retired concept, a FSN is no longer a unique concept identifier (I mean, theoretically, it would work if there's only the one retired concept with a FSN, but I imagine most times you're retiring a concept like that, it's to replace it with a different concept with the same FSN).

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,64 @@
...
+@Component
+public class HibernateInitializerDAO implements InitializerDAO {
+	
+	private static final Logger log = LoggerFactory.getLogger(HibernateInitializerDAO.class);
+	
+	@Autowired
+	private SessionFactory sessionFactory;
+	
+	@Override
+	public Concept getConceptByName(String name) {
```
ibacher, line 34: Shouldn't we take some account of locale here? It's completely possible for a concept to have the same FSN in both English and French, for example...

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,64 @@
...
+		
+		@SuppressWarnings("unchecked")
+		List<ConceptName> list = criteria.list();
+		
+		if (list.size() == 1) {
+			return list.iterator().next().getConcept();
+		} else if (list.isEmpty()) {
+			log.warn("No concept found for '" + name + "'");
+		} else {
+			throw new RuntimeException("Multiple concepts with the same fully specified name found for '" + name + "'");
```
ibacher, line 60: While these kinds of errors are probably OK in the API, in Iniz, this is actually a user-facing error message (since this error will be logged for the user and they'll have to figure out what is wrong), in which case, it would be nice to have some way of identifying _which_ concepts share a FSN so they can be fixed. This could be as simple as:

```suggestion
			throw new RuntimeException("Multiple concepts with the same fully specified name found for '" + name + "':\n" + list.stream().map(Concept::getUuid).collect(joining("\n")));
```

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
mseaton: No, the whole point is that we want this to be independent of Locale.  The use case is for uniquely identifying Concepts.  If more than one concept has the same FSN, then the name shouldn't be used as an identifier.   But actually, what we probably want to do is to have the DAO return a List<Concept> of all concepts that match the FSN (independent of Locale).  Then, the service should make the judgement around what to do with this List depending on the size of members returned (which is to fail on duplicate and return null on empty).  Make sense @Ruhanga

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
mseaton: We should also probably change this method name.  It isn't descriptive enough for what it is actually doing.  It should read something like "getUnretiredConceptByFullySpecifiedName" or "getUniqueConceptByFullySpecifiedName" or something.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
Ruhanga: Makes sense @mseaton. Will implement the [agreed on logic below](https://github.com/mekomsolutions/openmrs-module-initializer/pull/218#discussion_r1118869129).

api/src/test/java/org/openmrs/module/initializer/api/ConceptsLoaderIntegrationTest.java
Ruhanga: Ok..

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
Ruhanga: Resolved.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
Ruhanga: Ok..

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
Ruhanga: Done...

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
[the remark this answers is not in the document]
Ruhanga: Ok..

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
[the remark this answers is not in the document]
Ruhanga: Javadoc added.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,64 @@
...
+ * 
+ * @see InitializerConceptService
+ */
+@Component
+public class HibernateInitializerDAO implements InitializerDAO {
+	
+	private static final Logger log = LoggerFactory.getLogger(HibernateInitializerDAO.class);
+	
+	@Autowired
+	private SessionFactory sessionFactory;
```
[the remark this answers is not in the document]
Ruhanga: Done..

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * The Hibernate class for Concepts <br>
+ * <br>
+ * Use {@link InitializerService} to access these methods
+ * 
+ * @see InitializerConceptService
+ */
+@Component
+public class HibernateInitializerDAO implements InitializerDAO {
```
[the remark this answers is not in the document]
Ruhanga: Done...

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,64 @@
...
+import org.openmrs.ConceptName;
+import org.openmrs.api.ConceptNameType;
+import org.openmrs.api.context.Context;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * The Hibernate class for Concepts <br>
```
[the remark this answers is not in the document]
Ruhanga: Corrected, thanks for the review @mseaton.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,77 @@
...
+		if (list.size() == 1) {
+			return list.iterator().next().getConcept();
+		} else if (list.isEmpty()) {
+			log.warn("No concept found for '" + name + "'");
+		} else {
+			List<Concept> concepts = list.stream().map(ConceptName::getConcept).collect(Collectors.toList());
+			throw new RuntimeException("Multiple concepts with the same fully specified name found for '" + name + "':\n"
+			        + concepts.stream().map(Concept::getUuid).collect(Collectors.joining("\n")));
+		}
+		return null;
```
mseaton, line 75: So, what about the situation where a single concept has the same FSN in multiple languages?  For example, we might have the FSN "Tumor" in both English and Spanish.  In this case, we would not want this method to fail, since "Tumor" would refer to a single Concept.  Right, @Ruhanga  / @ibacher ?

Per my comment above, I think this DAO method should retrieve the distinct Concept projections that match the criteria query and return that List/Set of distinct matching Concepts.

So this method would return Set<Concept> (or  List<Concept> where we make sure there are no duplicates in the List)

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,77 @@
...
+	public void setSessionFactory(SessionFactory sessionFactory) {
+		this.sessionFactory = sessionFactory;
+	}
+	
+	/**
+	 * @see org.openmrs.module.initializer.api.InitializerService#getUnretiredConceptByFullySpecifiedName(String)
+	 */
+	@Override
+	public Concept getUnretiredConceptByFullySpecifiedName(String name) {
+		if (StringUtils.isBlank(name)) {
```
mseaton, line 46: Thinking this over, I think this DAO method would be more useful returning a List<Concept> or Set<Concept> and be named "getUnretiredConceptsByFullySpecifiedName(String name)", and then the consumer (either the service method or the consumer of that) would contain the logic to check whether there is 0, 1, or >1 result, and throw and exception if > 1.  That seems more appropriate to API logic, not DAO query logic.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -205,4 +217,12 @@ public Boolean getBooleanFromKey(String key) {
 	public InitializerConfig getInitializerConfig() {
 		return cfg;
 	}
+	
+	/**
+	 * @see org.openmrs.module.initializer.api.InitializerService#getUnretiredConceptByFullySpecifiedName(String)
+	 */
+	@Override
+	public Concept getUnretiredConceptByFullySpecifiedName(String name) {
+		return initializerDAO.getUnretiredConceptByFullySpecifiedName(name);
```
mseaton, line 226: Per my comment above, if a valid scenario exists where an FSN could be shared by multiple concepts, then this method should return a Set or List of those Concepts, and it should be down to the consuming class to figure out whether it can handle multiple concepts or not.

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -173,23 +174,8 @@ public static Concept fetchConcept(String id, ConceptService service) {
...
-					}
-					return concept;
-				}
-			}
-		}
-		finally {
-			Context.setLocale(originalLocale);
-		}
-		return null;
+		Concept concept = Context.getService(InitializerService.class).getUnretiredConceptByFullySpecifiedName(id);
```
mseaton, line 177: If the service and DAO methods change above to return a List<Concept>, _this_ is the appropriate place to check if the size of the returned List is 0 (return null), 1 (return the first member), or > 1 (throw an Exception that lists the offending concepts).

mseaton (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
ibacher: While I'm not opposed to returning a list or set from this method, to address the "Tumor / Tumor" thing, we just need to change the query so that its fetching concepts (instead of concept names) and do a unique projection... We'd still end up with a single concept.

The value to returning multiple concepts is if we expect more than one _concept_ to have the same FSN, but in different locales...

mseaton: Correct @ibacher - there are 2 separate things here.  Either way, we need to fix this method so that it is returning distinct projections of Concept, not just counting concept names.  *Then*, we need to decide separately if we want the logic as to whether to return one concept or a list, and where to throw exceptions to be in the DAO, service, or consumers of the service.

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
Ruhanga: I've added this change.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
Ruhanga: Method now returns List of Concepts.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
Ruhanga: Method now returns List of Concepts.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
Ruhanga: I've added projection on distinct concepts.

Ruhanga (commented):
Hi @ibacher, @mseaton, thanks for your reviews. I've addressed most of the requested changes. Could you have a look again? Thanks.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
```diff
@@ -0,0 +1,74 @@
...
+		
+		criteria.createAlias("concept", "concept");
+		criteria.add(Restrictions.eq("concept.retired", false));
+		criteria.setProjection(Projections.distinct(Projections.property("concept")));
+		
+		@SuppressWarnings("unchecked")
+		List<Concept> list = criteria.list();
+		
+		if (list.isEmpty()) {
+			log.warn("No concept found for '" + name + "'");
```
mseaton, line 70: One quick thing @Ruhanga  - I wonder if we can reduce the level of this warning.  Where this method is typically used, we are in a situation where we have a lookup string and are attempting to match on uuid, mapping, name, etc.  So the fact that no matching name is found when this method is called is not necessarily a warning-level event, and can cause unnecessary concern when looking at logs.  Thoughts?

api/src/main/java/org/openmrs/module/initializer/api/InitializerService.java
```diff
@@ -117,4 +117,12 @@ public interface InitializerService extends OpenmrsService {
 	 * @return the configuration of the Initializer module as defined by the runtime properties
 	 */
 	InitializerConfig getInitializerConfig();
+	
+	/**
+	 * Return a Concept that matches the exact fully specified name
```
mseaton, line 122: Minor comment that javadoc now needs to change to reflect return of multiple concepts

mseaton (approved):
Couple of quick things to fix but otherwise looks great.  Approving on my end.

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
Ruhanga: The warning-level IMO is fairly ok to maintain [from the previous implementation in core](https://github.com/openmrs/openmrs-core/blob/2.1.x/api/src/main/java/org/openmrs/api/db/hibernate/HibernateConceptDAO.java#L1739) since it'd ideally help identify a concept that'd be getting newly created and/or missing for any other domain requiring it via this method. This can be quite handy unlike having to go through a more detailed debug log that can be quite verbose to clearly zero-down to the cause.

ibacher: At the very least, this logging seems unnecessary here if we're also doing it in the Utils method no? I think we can leave out the logging here and allow the level above to log something... As written, this logs two warnings if the concept isn't found which seems like a lot.

ibacher: I also would tend to prefer this at the `info` level rather than `warn`.

api/src/main/java/org/openmrs/module/initializer/api/utils/Utils.java
```diff
@@ -229,21 +230,15 @@ public static Concept fetchConcept(String id, ConceptService service) {
...
-					return concept;
-				}
-			}
-		}
-		finally {
-			Context.setLocale(originalLocale);
+		
+		List<Concept> concepts = Context.getService(InitializerService.class).getUnretiredConceptsByFullySpecifiedName(id);
+		if (concepts.size() == 1) {
+			return concepts.iterator().next();
```
ibacher, line 236:
```suggestion
			return concepts.get(0);
```

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
Ruhanga: We've done away with the one in the DAO method @ibacher.

api/src/main/resources/moduleApplicationContext.xml
```diff
@@ -34,6 +37,9 @@
         </property>
         <property name="target">
             <bean class="org.openmrs.module.initializer.api.InitializerServiceImpl">
+            	<property name="initializerDAO">
```
ibacher, line 40: Minor, but could we consistently use spaces in this XML file?

api/src/main/java/org/openmrs/module/initializer/api/InitializerService.java
```diff
@@ -117,4 +117,12 @@ public interface InitializerService extends OpenmrsService {
 	 * @return the configuration of the Initializer module as defined by the runtime properties
 	 */
 	InitializerConfig getInitializerConfig();
+	
+	/**
+	 * Returns Unretired unique Concepts that match the exact fully specified name across all locales
```
ibacher, line 122:
```suggestion
	 * Returns non-retired unique concepts whose fully specified name exactly matches the supplied name in any locale
```

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
ibacher: This comment is on the DAO method, right? Because I still see it in this PR...

api/src/main/resources/moduleApplicationContext.xml
Ruhanga: Yes @ibacher, thanks.

ibacher (approved):

api/src/main/java/org/openmrs/module/initializer/api/HibernateInitializerDAO.java
Ruhanga: In the case that everything is running okay, it is ok to hide such logs unlike when something failed. I'd keep the logs maybe at INFO level like suggested.

Issue #160, opened 2021-11-11
Support for cohort types, cohort attribute types and cohort member attribute types.


mks-d:
[TBC] and also see https://github.com/openmrs/openmrs-module-cohort

rbuisson:
Hi @ibacher @mks-d , I see the need for Cohort Type, but I am not sure about Cohort Attribute Type and Cohort Member Attribute.
Is there a use case behind this? If so, how urgent?

ibacher:
Cohort Attribute Type (and hence Cohort Attributes) would be the obvious place to store the filters / criteria that define a a dynamic cohort.

I don't really have an obvious or immediate use-case for Cohort Member Attribute Type, but it's not likely to be very challenging to implement.

mks-d:
AFAIK the support for cohort-related metadata was developed and is sufficiently supported by Iniz.

PR #220, opened 2023-02-20
[#160] Add support for Cohort Types from the Cohort module

ibacher:
This is necessary for the Patient List feature in O3. Hopefully, the implementation is straight-forward enough, as this is a pretty trivial domain.


api/src/main/java/org/openmrs/module/initializer/api/cohort/ct/CohortTypeCsvParser.java
```diff
@@ -0,0 +1,76 @@
...
+import org.openmrs.module.cohort.api.CohortTypeService;
+import org.openmrs.module.initializer.Domain;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.openmrs.module.initializer.api.CsvParser;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component
```
Ruhanga, line 15: I'd remove `@Component ` in favour of `@OpenmrsProfile` here and on other occurrences.

pom.xml
```diff
@@ -166,6 +167,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.openmrs.module</groupId>
+            <artifactId>cohort-api</artifactId>
+            <version>${cohortModuleVersion}</version>
+            <scope>provided</scope>
+        </dependency>
+
```
Ruhanga, line 176: Let's go ahead to add this dependency to the validator so it can find the required classes when building it's application context.

Ruhanga (commented):
Thank you @ibacher. The PR is generally neat. Let's work on a few minor comments below.

api/src/main/java/org/openmrs/module/initializer/api/cohort/ct/CohortTypeCsvParser.java
Ruhanga: The challenge with this is that, 1) the Spring Framework tends to consider `@Component` which overrides `@OpenmrsProfile` and so will always attempt loading the bean irrespective of the conditions specified on the `@OpenmrsProfile` annotation. 2) minor though, it's kind of semantically not okay IMO for a duplication of effort in using more than one annotation doing the same thing.

mks-d: @ibacher, admittedly this is old in my mind, but I also remember that `@Component` and `@OpenmrsProfile` do not bode well together, or at least they did not at some point in the past.

Now if you confirm that it is 100% ok and that it is harmless to leave `@Component` in, then fine.

ibacher: I wasn't aware that it caused problems, but actually, I think I see where this might arise, and also might help explain why using annotations can perform so terribly in OMRS:

1. In Core, we have a [component-scan](https://github.com/openmrs/openmrs-core/blob/acfaa3559a49bd3c59f29cd761f02da2fc388a82/api/src/main/resources/applicationContext-service.xml#L671) defined for `org.openmrs` (which will also recursively scan... `org.openmrs.**.*`). On that component-scan, we define an `exclude-filter`, which actually utilises the check on the `@OpenmrsProfile` annotation.
2. In modules, like Iniz, we tend to add a [component-scan](https://github.com/mekomsolutions/openmrs-module-initializer/blob/41a275cc75040c23de34ab7537628b2d5e232708/api/src/main/resources/moduleApplicationContext.xml#L26) without the same filters.

I guess what that means then is that a bean might not be created as part of the initial scan, but  could be loaded (because it's annotated with `@Component`) by the second context-scan.

ibacher, line 15:
```suggestion
```

api/src/main/java/org/openmrs/module/initializer/api/cohort/ct/CohortTypeLineProcessor.java
```diff
@@ -0,0 +1,26 @@
+package org.openmrs.module.initializer.api.cohort.ct;
+
+import org.apache.commons.lang3.StringUtils;
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.cohort.CohortType;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.springframework.stereotype.Component;
+
+@Component
```
ibacher, line 10:
```suggestion
```

api/src/main/java/org/openmrs/module/initializer/api/cohort/ct/CohortTypeLoader.java
```diff
@@ -0,0 +1,17 @@
+package org.openmrs.module.initializer.api.cohort.ct;
+
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.cohort.CohortType;
+import org.openmrs.module.initializer.api.loaders.BaseCsvLoader;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
```
ibacher, line 9:
```suggestion
```

ibacher:
Alright... having addressed those issues, what's left to do here?

ibacher:
Alright... Anything else that needs to be done here?

mks-d (approved):

PR #222, opened 2023-02-24
Fix FHIR domains to properly use @OpenmrsProfile

ibacher:
Per [this insight](https://github.com/mekomsolutions/openmrs-module-initializer/pull/220#discussion_r1114095110), removing `@Component` from some places it shouldn't be.


Ruhanga (approved):
LGTM, thanks @ibacher.

Issue #219, opened 2023-02-17
Add Support for Drug Reference Maps


mogoodrich:
Drugs in OpenMRS can be mapped to reference terms in the same manner as Concepts (see Drug Reference Map in orders)

We should support this in Iniz as follows:

In the Drugs domain, support specifying mapping in the same was we do for concepts as described here:

https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/concepts.md   (see "Mapping Headers" section... we will support the new recommended format "Mappings|SAME-AS" aod "Mappings|SAME-AS|Source" format)

@mks-d  @ibacher @mseaton @rbuisson  fyi... any thoughts? At PIH we need this and can likely build this out in the coming weeks.

mks-d:
Thanks @mogoodrich - I will let others comment in more details but if this is
1. Useful to PIH
2. Backward compatible (looks like it would be)
3. Documented in the README(s)

then we are welcoming this suggested improvement. Depending on the timing this would make it in 2.5.0 or 2.6.0.

ibacher:
This seems like an excellent addition to me!

mogoodrich:
Yes, @mks-d  all three of those things will be true, thanks!

mseaton:
I ran into a bit of a snag here.  See https://issues.openmrs.org/browse/TRUNK-6164.  Will need to discuss how to work around this in Initializer.  @ibacher  FYI

mseaton:
I have a solution for this ready for review @ibacher .  Unfortunately the PR is failing in Github Actions because it depends on the latest 2.3.x build, and OpenMRS CI is failing to build 2.3.x due to Bamboo issues.  Once we resolve that, things should turn green.

PR #217, opened 2023-02-08
[#209] Fixing Validator.

Ruhanga:
Issue: https://github.com/mekomsolutions/openmrs-module-initializer/issues/209

Depends on https://github.com/openmrs/openmrs-core/pull/4250

PR #226, opened 2023-03-02
[#219] Add drug mapping line processor.

mseaton:
This still needs updates to the README and release notes, and the unit test changes to the `DrugsLoaderIntegrationTest` fails due to the dependency on versions of core that does not have the bug fix I added. Those need to be figured out.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/drugs/MappingsDrugLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/drugs/MappingsDrugLineProcessor.java
@@ -1,0 +1,111 @@
+package org.openmrs.module.initializer.api.drugs;
+
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.openmrs.ConceptMapType;
+import org.openmrs.ConceptReferenceTerm;
+import org.openmrs.ConceptSource;
+import org.openmrs.Drug;
+import org.openmrs.DrugReferenceMap;
+import org.openmrs.api.ConceptService;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.openmrs.module.initializer.api.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component("initializer.mappingsDrugLineProcessor")
+public class MappingsDrugLineProcessor extends DrugLineProcessor {
+	
+	protected final Logger log = LoggerFactory.getLogger(getClass());
+	
+	public static final String MAPPING_HEADER_PREFIX = "mappings";
+	
+	public static final String MAPPING_HEADER_SEPARATOR_REGEX = "\\|";
+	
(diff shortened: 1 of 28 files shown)


mks-d (commented):
I leave it to @ibacher to review the code, I'll review the documentation when ready. Thanks a bunch @mseaton 🙏

mseaton:
@ibacher - I took your advice and put the tests related to the drug reference maps into the 2.3 submodule, and then made this submodule use the latest 2.3.6-SNAPSHOT version, rather than 2.3.0.  I used 2.3 as this is the furthest back I backported this fix into core.  This shouldn't impact validator, as validator hard-codes a specific 2.3.x version into it's own pom.

api/src/main/java/org/openmrs/module/initializer/api/drugs/MappingsDrugLineProcessor.java
```diff
@@ -0,0 +1,111 @@
...
+import org.openmrs.module.initializer.api.CsvLine;
+import org.openmrs.module.initializer.api.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component("initializer.mappingsDrugLineProcessor")
+public class MappingsDrugLineProcessor extends DrugLineProcessor {
```
ibacher, line 21: Is there a reason this extends `DrugLineProcessor`?

ibacher (approved):
Aside from one weird bit of inheritance, this LGTM!

api/src/main/java/org/openmrs/module/initializer/api/drugs/MappingsDrugLineProcessor.java
mseaton: @ibacher - good question.  I'm not sure.  I followed the same pattern from the Concepts domain, since this is an analog to what is done there with mappings on concepts.  In that domain, all of the various LineProcessors seem to extend ConceptLineProcessor, so I stuck with that.  I can just make it extend BaseLineProcessor<Drug> instead.  I'm not particular, I just didn't know if the Iniz framework had a reason to do it this way, and following an established pattern seemed safest.

ibacher:
@mseaton Any idea on the test failures? Looks disconnected, but I'm worried that we've added something strange here...

Ruhanga (commented):
Hi @ibacher, @mseaton, not sure if the failing tests are of a time-bomb nature but, locally they pass when build against `openjdk version "1.8.0_265"` but fail similarly on standard Java 8 (I'm not sure which JVM flavour Github actions is using). I'm also suprised how these failing tests survived in the previous commits despite the commit changes touching the tests logic. I've somewhat addressed the failing tests in the last commit on this PR: https://github.com/mekomsolutions/openmrs-module-initializer/pull/217.

mseaton:
Thanks @Ruhanga !  That's very odd, because it looks like github actions is using OpenJDK 1.8.362.  I went so far as to create a vanilla Ubuntu 22.04 Docker container, install this exact temurin Java version, and build, and I was able to do so successfully, so I've been at a real loss as to what is going on in Github actions.

My suspicion has been that it isn't directly related to this PR/commit, but I haven't had any evidence to that.  Are you saying that these errors are already existing in the codebase in the master branch and you have some fixes in place in another PR that will address them?

PR #228, opened 2023-03-13
[#211] Undo InitializerDAO XML injection in favour of autowiring injection.

Ruhanga:
This is a followup PR to address a runtime bug caused by the InitializerDAO not being set.
Issue: https://github.com/mekomsolutions/openmrs-module-initializer/issues/211

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
@@ -60,6 +60,7 @@ public class InitializerServiceImpl extends BaseOpenmrsService implements Initia
 	 * 
 	 * @param initializerDAO The data access object to use
 	 */
+	@Autowired
 	public void setInitializerDAO(InitializerDAO initializerDAO) {
 		this.initializerDAO = initializerDAO;
 	}
--- a/api/src/main/resources/moduleApplicationContext.xml
+++ b/api/src/main/resources/moduleApplicationContext.xml
@@ -36,11 +36,7 @@
             <ref bean="transactionManager" />
         </property>
         <property name="target">
-            <bean class="org.openmrs.module.initializer.api.InitializerServiceImpl">
-                <property name="initializerDAO">
-                    <ref bean="initializerDAO" />
-                </property>
-            </bean>
+            <bean class="org.openmrs.module.initializer.api.InitializerServiceImpl"/>   
         </property>
         <property name="preInterceptors">
             <ref bean="serviceInterceptors" />


mks-d (approved):
If this solves a bug at runtime, then it's good to go for me.

api-2.3/src/test/java/org/openmrs/module/initializer/api/DrugsLoaderWithMappingsIntegrationTest.java
```diff
@@ -34,6 +34,11 @@ public class DrugsLoaderWithMappingsIntegrationTest extends DomainBaseModuleCont
 	@Autowired
 	private DrugsLoader loader;
 	
+	@Override
+	public void updateSearchIndex() {
+		// to prevent Data Filter's 'Illegal Record Access'
```
mks-d, line 39: @Ruhanga do you know why Data Filter interferes with this new domain?

mks-d: Looks like all api-2.3 (and beyond) Spring tests will need this because Data Filter becomes a dependency from 2.3, so let's make sure that this becomes the default behaviour for all tests in api-2.3.

Let's make it its own PR.

mks-d (changes requested):

api-2.3/src/test/java/org/openmrs/module/initializer/api/DrugsLoaderWithMappingsIntegrationTest.java
Ruhanga: PR [here](https://github.com/mekomsolutions/openmrs-module-initializer/pull/229) @mks-d.

mks-d (approved):

mseaton:
@Ruhanga / @mks-d .  This was already merged, so not sure you need or want my review still.  No reason to oppose this, as long as it works and doesn't lead to performance issues, particularly at startup.  I _would_ like to know why this is necessary though - do we know what kind of situation works when we inject dependencies via autowiring and does not work when we explicitly inject using xml mappings?  Seems odd.

mogoodrich:
@Ruhanga  @mks-d  I confirmed that this change (Autowired in a service) literally added 4 minutes to the start up time of our PIH EMR dev environments (haven't checked production instances but I assume the same as well). 

This is due to this pesky issue that we've never figured out how to solve besides not autowiring in bean.

https://issues.openmrs.org/browse/TRUNK-3806

This is a blocker for our development, can we revert this until we figure it out, or at least revert and do a release before adding, because we need @mseaton 's drug CSV enhancement that only in this snapshot.  Thanks!

Issue #231, opened 2023-03-23
Initializer to build against Java 11


Ruhanga:
### Background
Currently Initializer can only build successfully against major Java version 8 and 11.
### Acceptance Criteria
* Replace deprecated libraries from Java 8 with proper alternatives.
* Ensure no test failures.
* Expand GitHub Actions to ensure that each PR's build is made against the following versions of Java:
   * Java 8
   * Java 11

Ruhanga:
Closing this as resolved.

PR #232, opened 2023-03-24
[#231] Initializer to build against Java 11

Ruhanga:
Issue #231


mks-d (commented):
I will defer the review to @ibacher.

ibacher:
Other than the fact that I'd probably have written `matrix.java` rather than `matrix.Java`, this looks great to me!

mks-d (approved):

PR #245, opened 2023-07-13
Add support for Cohort Attribute types

icrc-loliveira:
## Domain 'cohortattributetypes'

The **cohortattributetypes** subfolder contains CSV import files for defining cohort Attribute types, used by the cohort module.

This is a possible example of its contents:
```bash
cohorttypes/
  ├──cohortattributetypes.csv
  └── ...
```

The format of this CSV should be as follows:

| <sub>Uuid</sub> |<sub>Void/Retire</sub> | <sub>Name</sub> | <sub>Description</sub> | <sub>Datatype classname</sub> | <sub>Min occurs</sub> | <sub>Max occurs</sub> | <sub>Preferred handler classname</sub> | <sub>Handler config</sub> | <sub>_order:1000</sub> |
| - | - | - | - | - | - | - | - | - | - |
| <sub>09790099-9190-429d-811a-aac9edb8d98e</sub> | | <sub>Control</sub> | <sub>This is a control group.</sub> | <sub>org.openmrs.customdatatype.datatype.FreeTextDatatype</sub> |

Headers that start with an underscore such as `_order:1000` are metadata headers. The values in the columns under those headers are never read by the CSV parser.

###### Header `Name`

This is *required* for every entry. This will be the name of the cohort attribute type, wherever that is used.

###### Header `Description`

This is *required* for every entry. This will be the description of the cohort attribute type, wherever that is used.

###### Header `Datatype classname` *(mandatory)*
This points to the Java type of the attribute that is defined by the attribute type. [Here](https://github.com/openmrs/openmrs-core/tree/2.1.1/api/src/main/java/org/openmrs/customdatatype/datatype) is the list of possible values.


api/src/main/java/org/openmrs/module/initializer/api/cohort/cat/CohortAttributeTypeLineProcessor.java
```diff
@@ -0,0 +1,24 @@
+package org.openmrs.module.initializer.api.cohort.cat;
+
+import org.apache.commons.lang3.StringUtils;
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.cohort.CohortAttributeType;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+
+@OpenmrsProfile(modules = { "cohort:3.2.* - 9.*" })
+public class CohortAttributeTypeLineProcessor extends BaseLineProcessor<CohortAttributeType> {
```
ibacher, line 10: It should be possible to extend `BaseAttributeTypeLineProcessor`. The other properties that provides will be useful here as well (`datatypeClassname`, `datatypeConfig`, `minOccurs`, `maxOccurs`, `preferredHandlerClassname`, and `handlerConfig`).

I think `datatypeClassname` is actually required.

ibacher (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/cohort/cat/CohortAttributeTypeLineProcessor.java
```diff
@@ -0,0 +1,29 @@
...
+		String uuid = line.getUuid();
+		if (StringUtils.isNotBlank(uuid)) {
+			instance.setUuid(line.getUuid());
+		}
+		
+		instance.setName(line.getName(true));
+		instance.setDescription(line.get("Description"));
+		instance.setDatatypeConfig(line.get("Datatype classname", true));
+		instance.setDatatypeConfig(line.get("Preferred handler classname"));
+		instance.setDatatypeConfig(line.get("Handler config"));
```
ibacher, line 23: These should be three separate properties, right?

I think it would be easier to make this class extend `BaseAttributeTypeLineProcessor` and not have to define this method at all (since there's nothing particularly distinct about the lines that we process).

ibacher (approved):

Issue #248, opened 2023-10-05
Support setting names and descriptions of metadata from translations in the context locale


mseaton:
Currently, we have a setup where we have a single master set of metadata files (eg. a csv of encounter types) that is shared across multiple countries.  It is often requested that in each country, the actual name attached to the metadata (eg. encounter_type.name) be in the default locale used in that country.  This is useful, for example, when reports are run directly against the database, and the encounter_type.name is pulled in.  It is also useful if there are areas of the UI that do not use the localized display correctly.

One idea is to allow one to specify that the name and/or description of a given metadata row should be set based on the display in the default context locale.

mks-d:
@mseaton is this still current or we can close this issue?

mseaton:
There does appear to be an open PR around this @mks-d .  I can't remember the status:
https://github.com/mekomsolutions/openmrs-module-initializer/pull/249

mks-d:
Yes 😞, I left a comment there too → https://github.com/mekomsolutions/openmrs-module-initializer/pull/249#discussion_r1658869248

PR #249, opened 2023-10-05
#248 - Support setting names and descriptions of metadata from transl…

mseaton:
…ations in the context locale

Note, if this approach is approved, then there are likely additional changes to expand this beyond the encounter type domain, by changing other metadata loaders to extend the BaseMetadataLineProcessor rather than the BaseLineProcessor and then update the way the set the name and description properties.

For now, this works and is tested against the EncounterType domain.

We'll also need to update the README and Release Notes.  I can add that in if this is agreed on as a general approach.  I expect there may be alternative ideas for what the constant is named (currently "_translate_", but I assume others may have better ideas here).

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/BaseMetadataLineProcessor.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/BaseMetadataLineProcessor.java
@@ -1,0 +1,46 @@
+package org.openmrs.module.initializer.api;
+
+import org.openmrs.OpenmrsMetadata;
+import org.openmrs.messagesource.MessageSourceService;
+
+/**
+ * Extends the BaseLineProcessor by adding methods to retrieve name and description based
+ * on the configured name and description and translations of these if appropriate.
+ */
+public abstract class BaseMetadataLineProcessor<T extends OpenmrsMetadata> extends BaseLineProcessor<T> {
+	
+	public static final String TRANSLATE = "_translate_";
+	
+	protected MessageSourceService messageSourceService;
+	
+	public BaseMetadataLineProcessor(MessageSourceService messageSourceService) {
+		super();
+		this.messageSourceService = messageSourceService;
+	}
+	
+	/**
+	 * @return the name defined in the current line, which can indicate if it should come from a message
+	 *         code
+	 */
+	public String getName(T metadata, CsvLine line) {
+		String name = line.getName(true);
+		if (TRANSLATE.equals(name)) {
+			String code = "ui.i18n." + metadata.getClass().getSimpleName() + ".name." + metadata.getUuid();
+			name = messageSourceService.getMessage(code);
+		}
+		return name;
+	}
+	
+	/**
+	 * @return the description defined in the current line, which can indicate if it should come from a
+	 *         message code
+	 */
+	public String getDescription(T metadata, CsvLine line) {
+		String description = line.get(HEADER_DESC);
+		if (TRANSLATE.equals(description)) {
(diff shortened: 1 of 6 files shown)


mks-d (commented):
@mseaton I'm surely a bit out of context but is this for metadata entities that do not support a display string (for which we would use the `display:xy` header, like for example [here](https://github.com/openmrs/ozone-distro-cambodia/blob/a5ddcb3999d30e04c8ff89efd736710dd6ce276c/configs/openmrs_config/visittypes/visittypes.csv#L2))?

mks-d (commented):
Ok sorry, just read the issue [description](https://github.com/mekomsolutions/openmrs-module-initializer/issues/248#issue-1928695972)...

So, at Iniz-time, the metadata name and description are localised as per the system's locale, correct?

This should not have any impact in the UX (in theory), but is useful in the local context? As you describe when flattening data directly out of the database, etc?

If that's right then this really a metadata regionalization/localization feature so that the database looks "local" to consumers in country. Then maybe, rather than the `_translate_` directive, we could use `_localize_` (or `_L10n_`)?

mseaton:
> If that's right then this really a metadata regionalization/localization feature so that the database looks "local" to consumers in country. Then maybe, rather than the `_translate_` directive, we could use `_localize_` (or `_L10n_`)?

Sure, we can have this named whatever we want.  I'll take direction from you on this, and make it so.

api/src/main/java/org/openmrs/module/initializer/api/BaseMetadataLineProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+	
+	/**
+	 * @return the name defined in the current line, which can indicate if it should come from a message
+	 *         code
+	 */
+	public String getName(T metadata, CsvLine line) {
+		String name = line.getName(true);
+		if (TRANSLATE.equals(name)) {
+			String code = "ui.i18n." + metadata.getClass().getSimpleName() + ".name." + metadata.getUuid();
+			name = messageSourceService.getMessage(code);
```
mks-d, line 29: What locale is going to be used here? My question is at a functional level, OpenMRS' default locale I would assume?

mseaton: Based on the functioning of the unit tests, it is whatever is returned from Context.getLocale(), which will end up as LocaleUtility.getDefaultLocale().  This will ultimately be whatever is in the GP named "default_locale", and if this is either unset or can't be retrieved, will fall back to "en_GB"

mks-d: Does that describe the feature appropriately:

>Support for loading-time L10n of metadata names and descriptions based on Context's locale.

?
It would need to be documented well of course, but is that functional intent?

mseaton: Yes, but it raises a good point.  We might want to change this and also be more explicit in the code.  Because if someone were to decide to re-run an initializer load themselves (eg. we have an admin feature to allow this for development), then this will pick up their locale, rather than a standard system locale.  So maybe the translation here should explicitly pass in the locale that is returned from LocaleUtility.getDefaultLocale(), so that this will always behave consistently in the same system.  Then, the documentation can describe this as "...based on the system default locale (default_locale global property)" or something.

mks-d: @mseaton we all left this PR and thread hanging, what's your take? Should we resume reviewing it, or should be close it for now?

PR #253, opened 2023-10-20
Add Domain for FhirContactPoints

mherman22:


api-2.5/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapCsvParser.java
```diff
@@ -0,0 +1,140 @@
...
+				baseAttributeType = locationService.getLocationAttributeTypeByName(attributeType);
+				
+				if (baseAttributeType != null) {
+					return baseAttributeType;
+				}
+				
+				return locationService.getLocationAttributeTypeByUuid(attributeType);
+				break;
+			case PROVIDER:
+				baseAttributeType = providerService.getProviderAttributeTypeByName(attributeType);
```
mherman22, line 129: i guess we need to backport [this commit](https://github.com/openmrs/openmrs-core/commit/eadc4388ed912339b60713c8daa33f59ba5bcb19) (that introduces the `getProviderAttributeTypeByName()` method) to the 2.5.x branch curently at 5.13.0-SNAPSHOT branch.

cc: @ibacher

ibacher: Hmmm... let's just target this at 2.6+ then. I think that's less confusing.

mherman22: Awesome,lemi update that

api/src/test/java/org/openmrs/module/initializer/api/FhirPatientContactPointMapIntegrationTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.openmrs.module.initializer.api.fhir.cpm.FhirContactPointMapLoader;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.Optional;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+public class FhirPatientContactPointMapIntegrationTest extends DomainBaseModuleContextSensitiveTest {
```
ibacher, line 25: I think this test needs to be in the api-2.5 (probably api-2.6) submodule or something.

mherman22: done!

api-2.5/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapCsvParser.java
mherman22: done!

ibacher:
@mherman22 Is there no way to get this to work without shifting to 2.0.0 of the FHIR2 module? I would prefer to support older versions.

mherman22:
> @mherman22 Is there no way to get this to work without shifting to 2.0.0 of the FHIR2 module? I would prefer to support older versions.

I was trying this out. I'm pushing it back to 1.11.0 however the challenge there has to do with the location type translator not being got as a bean i.e NoSuchBeanException

omod/src/main/resources/config.xml
```diff
@@ -17,7 +17,11 @@
 	<conditionalResources>
 		<conditionalResource>
 			<path>/lib/initializer-api-2.4-${project.version}.jar</path>
-			<openmrsVersion>2.4.* - 2.*</openmrsVersion>
+			<openmrsVersion>2.4.* - 9.*</openmrsVersion>
+		</conditionalResource>
+		<conditionalResource>
+			<path>/lib/initializer-api-2.5-${project.version}.jar</path>
+			<openmrsVersion>2.5.13 - 9.*</openmrsVersion>
```
Ruhanga, line 24: I'd keep the upper version limit to `2.*` as major version changes are likely to package non backwards compatible changes.

ibacher: @Ruhanga It's still better to discover the errors in practice than to need to figure out why functionality is missing / fails by default. I.e., I think it's better to assume it will work and then deal with errors rather than pre-emptively exclude features.

ibacher:
It looks like we're going to need to make 2.6 the minimum version for the LoaderTest and probably the validator too for this to work. @Ruhanga Do you have an opinion on that? Otherwise, we maybe add something to the Loader to not try to run a domain if a valid version of a loader isn't available.

mherman22:
> It looks like we're going to need to make 2.6 the minimum version for the LoaderTest and probably the validator too for this to work. @Ruhanga Do you have an opinion on that? Otherwise, we maybe add something to the Loader to not try to run a domain if a valid version of a loader isn't available.

Fyi, I have done this locally though it is now not loading the loader for **QUEUE** so I am still looking for a workaround around that.

Ruhanga:
>It looks like we're going to need to make 2.6 the minimum version for the LoaderTest and probably the validator too for this to work.

@ibacher, @mherman22 concerning the above, upgrading to OpenMRS Core greater than 2.4.0 might require refactoring on the part of the Validator to accommodate new changes + upgraded libraries on Core. This will consequently need some extensive testing of the Validator itself.

ibacher:
First bit looks like we need to change the calls like:

```java
executeChangelog("changelog", null);
```

To:

```java
executeChangelog("changelog", (ChangeSetExecutorCallback) null);
```

mks-d (commented):
@mherman22 could you provide a detailed description to this PR so as to provide more context? Or @ibacher could you speak to it?

api-2.6/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapCsvParser.java
```diff
@@ -0,0 +1,147 @@
...
+import org.openmrs.module.initializer.Domain;
+import org.openmrs.module.fhir2.model.FhirContactPointMap;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.openmrs.module.initializer.api.CsvParser;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+@OpenmrsProfile(modules = { "fhir2:1.11.* - 9.*" }, openmrsPlatformVersion = "2.6.3 - 2.6.*, 2.7.* - 9.*")
```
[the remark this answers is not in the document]
mks-d: @ibacher is this (namely → `modules = { "fhir2:1.11.* - 9.*" }`) used because there is no way to say "`1.11.*` and above"?

ibacher:
@mks-d This PR adds support for a feature that allows mapping between OMRS Attributes and FHIR [ContactPoints](https://www.hl7.org/fhir/R4/datatypes.html#ContactPoint). It populates a table that we use in the FHIR2 module to map between them. So, if you have an attribute that represents something covered by FHIR's ContactPoints type (phone number, email, website, etc.) as either a PersonAttribute or standard OMRS attribute, it will show up in the FHIR2 output for Patient, Provider, Location, etc. in the appropriate semantic field rather than as an OMRS-specific extension.

api-2.6/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapCsvParser.java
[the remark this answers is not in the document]
ibacher: Actually, I think we could just use `fhir2:1.11.0` and possibly the same for `openmrsPlatformVersion`. I always forget how these get parsed.

README.md
```diff
@@ -149,6 +150,7 @@ This is the list of currently supported domains in their loading order:
 1. [Cohort Attribute Types (CSV files)](readme/cohort.md#domain-cohortattributetypes)
 1. [FHIR Concept Sources (CSV files)](readme/fhir.md#domain-fhirconceptsources)
 1. [FHIR Patient Identifier Systems (CSV Files)](readme/fhir.md#domain-fhirpatientidentifiersystems)
+2. [FHIR ContactPointMap (CSV Files)](https://github.com/mherman22/openmrs-module-initializer/blob/ContactPointMap-impl/readme/fhir.md#domain-fhircontactpointmap)
```
mks-d, line 153: The path is wrong here.

Also let's use normal Title Case: "Contact Point Map".

mks-d (changes requested):
@ibacher will you re-review this then?

README.md
```diff
@@ -60,6 +60,7 @@ configuration/
   ├── encounterroles/
   ├── fhirconceptsources/
   ├── fhirpatientidentifiersystems/
+  ├── fhirContactPointMap/
```
ibacher, line 63:
```suggestion
  ├── fhircontactpointmaps/
```

README.md
ibacher:
```suggestion
2. [FHIR Contact Point Map (CSV Files)](readme/fhir.md#domain-fhircontactpointmap)
```

api-2.6/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapCsvParser.java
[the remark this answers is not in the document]
ibacher: Let's try these, here and elsewhere:

```suggestion
@OpenmrsProfile(modules = { "fhir2:1.11.0" }, openmrsPlatformVersion = "2.6.3")
```

api-2.6/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapLineProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.fhir2.model.FhirContactPointMap;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.springframework.stereotype.Component;
+
+import static org.openmrs.module.initializer.api.fhir.cpm.FhirContactPointMapCsvParser.ATTRIBUTE_TYPE_DOMAIN_HEADER;
+
+@Component
+@OpenmrsProfile(modules = { "fhir2:1.11.* - 9.*" }, openmrsPlatformVersion = "2.6.3 - 2.6.*, 2.7.* - 9.*")
```
ibacher, line 23: See previous

api-2.6/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapLineProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+public class FhirContactPointMapLineProcessor extends BaseLineProcessor<FhirContactPointMap> {
+	
+	private static final String SYSTEM_HEADER = "system";
+	
+	private static final String USE_HEADER = "use";
+	
+	private static final String RANK_HEADER = "rank";
```
ibacher, line 30: I think headers are normally capitalized, e.g. "System", etc.

api-2.6/src/main/java/org/openmrs/module/initializer/api/fhir/cpm/FhirContactPointMapCsvParser.java
```diff
@@ -0,0 +1,146 @@
...
+public class FhirContactPointMapCsvParser extends CsvParser<FhirContactPointMap, BaseLineProcessor<FhirContactPointMap>> {
+	
+	public static final String ATTRIBUTE_TYPE_DOMAIN_HEADER = "Entity name";
+	
+	public static final String ATTRIBUTE_TYPE = "Attribute Type";
```
ibacher, line 33: Headers at least should be like "Entity name", so, e.g., "Attribute type", etc.

readme/fhir.md
```diff
@@ -73,4 +73,41 @@ This is the URL of the code system in FHIR. For terminologies identified
...
+fhirContactPointMap/
+  ├──contactPointMap.csv
+  └── ...
+```
+The format of this CSV should be as follows:
+
+| Uuid                                 | Void/Retire | Entity name | Attribute Type                       | System | Use  | Rank | _order:1000 |
+|--------------------------------------|-------------|-------------|--------------------------------------|--------|------|------|-------------|
+| fa48acc4-ef1f-46d6-b0af-150b00ddee9d |             | person      | 717ec942-3c4a-11ea-b024-ffc81a23382e | PHONE  | WORK | 1    |             |
+|                                      |             | person      | PAT_RENAME_NEW_NAME                  | PHONE  | HOME |      |             |
```
ibacher, line 94: Is the `Attribute Type` field here correct?

ibacher (changes requested):
@mherman22 Are you able to wrap up the necessary changes here?

mherman22:
> @mherman22 Are you able to wrap up the necessary changes here?

On it

api-2.6/pom.xml
```diff
@@ -0,0 +1,136 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <parent>
+        <groupId>org.openmrs.module</groupId>
+        <artifactId>initializer</artifactId>
+        <version>2.7.0-SNAPSHOT</version>
```
ibacher, line 9: We should update this now that 2.7.0 is released. I guess this should be 2.8.0-SNAPSHOT.

Issue #252, opened 2023-10-16
Support explicit `addresshierarchy` domain.


mks-d:
Introduce an `addresshierarchy` (AH) domain that takes care of:
1. Load all AH-related configs (address template, address entries, etc).
1. Filling the Ext I18n's i18n reverse cache.

This work will demand for Iniz to become aware of AH:
* Let's release AH 2.15.3 and bump it to AH 3.0.0.
  * Iniz will depend and be aware of AH 3.0.0.
* AH 3.0.0 will not support self-loading of AH configs anymore.
  * Let's create an ADDR ticket for this ↑ non-backward compatible change.

mseaton:
If we are removing a feature like this from AH, then we should bump the version to 3.0.0, not just up to 2.16.0.

mks-d:
Thanks @mseaton and for [the exchange on Slack](https://openmrs.slack.com/archives/C021QUDSD5K/p1697473045708299), I have updated the description accordingly.

Ruhanga:
I've created [ADDR-129](https://issues.openmrs.org/browse/ADDR-129) to followup the necessary refactoring on the AH module.

mks-d:
@Ruhanga was this not completed in the end?

PR #254, opened 2023-10-31
[#252] Support for 'addresshierarchy' domain.

Ruhanga:
#252


api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
```diff
@@ -0,0 +1,45 @@
...
+	public ConfigDirUtil getDirUtil() {
+		return new ConfigDirUtil(iniz.getConfigDirPath(), iniz.getChecksumsDirPath(), getDomainName(), true);
+	}
+	
+	@Override
+	public void loadUnsafe(List<String> wildcardExclusions, boolean doThrow) throws Exception {
+		
+		try {
+			AddressConfigurationLoader.loadAddressConfiguration(Paths.get(iniz.getConfigDirPath()), Paths.get(iniz.getChecksumsDirPath()), getDomain().toString());
+			Context.getService(AddressHierarchyService.class).initI18nCache();
```
Ruhanga, line 35: @mogoodrich this is always gona initialize the i18n cache at startup from here. `initializeFullAddressCache()` is implicitly called by `AddressConfigurationLoader.loadAddressConfiguration(...)`

Ruhanga: I've also tweaked this to ensure the `initializeFullAddressCache` is always called not only when checksums are changed.

mseaton (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
mogoodrich: Yes, @mseaton  that is my preference as well, that Iniz sticks to loading metadata, and *not* being responsible for setting up in-memory caches.  

I believe the original loadAddressConfiguration method already did #1: reloads the cache if the AH entries are reloaded.  #2 could be added as well, but is not really necessary if the if we keep the method that refreshes the cache in the contextRefreshed method of the Address Hierarchy activator.

@Ruhanga  @mks-d  I assume the intent of Iniz is primarily to do metadata loading, not to do overall module start-up tasks?  For instance, if for some reason Iniz was uninstalled from an implementation, I'd still expect a module to start up properly, as long as it's metadata was already loaded?

ibacher: > I assume the intent of Iniz is primarily to do metadata loading, not to do overall module start-up tasks?

Yes, I think that's explicitly the purpose here.

api/src/main/java/org/openmrs/module/initializer/Domain.java
```diff
@@ -21,6 +21,7 @@ public enum Domain {
 	PROVIDER_ROLES,
 	LOCATIONS,
 	LOCATION_TAG_MAPS,
+	ADDRESS_HIERARCHY,
```
ibacher, line 24: Should the address hierarchy be loaded before locations?

api/src/test/java/org/openmrs/module/initializer/api/AddressHierarchyLoaderIntegrationTest.java
```diff
@@ -54,14 +58,17 @@ public class AddressHierarchyMessagesLoadingTest extends DomainBaseModuleContext
...
 	@Before
 	public void setup() {
 		// Disabling AH full caching otherwise loading takes too long
 		Context.getAdministrationService().saveGlobalProperty(new GlobalProperty(
 		        AddressHierarchyConstants.GLOBAL_PROP_INITIALIZE_ADDRESS_HIERARCHY_CACHE_ON_STARTUP, "false"));
 		
-		Context.getAdministrationService()
-		        .saveGlobalProperty(new GlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST, "en, km_KH"));
+		Context.getAdministrationService().saveGlobalProperty(
+		    new GlobalProperty(OpenmrsConstants.GLOBAL_PROPERTY_LOCALE_ALLOWED_LIST, "en_GB, km_KH"));
```
ibacher, line 71: Is there a reason that this switched from `en` to `en_GB`? Does `en` not work? (Most locales in OpenMRS are stored as language without country specifier).

api/src/main/java/org/openmrs/module/initializer/Domain.java
Ruhanga: Address hierarchy and OpenMRS Location are different and/or disjoint entities, but sounds logical to load the `ADDRESS_HIERARCHY` domain after the `LOCATIONS` domain.

api/src/test/java/org/openmrs/module/initializer/api/AddressHierarchyLoaderIntegrationTest.java
Ruhanga: I was trying to add `km_KH` to the list, but looks like in tests too, `en_GB` is the [default locale](https://github.com/openmrs/openmrs-core/blob/c6de9235691e83dde50c047be9b2168eac1faf3b/api/src/main/java/org/openmrs/util/OpenmrsConstants.java#L328C29-L328C73) which constrains me to use it instead when populating the locale allowed list.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
Ruhanga: >For instance, if for some reason Iniz was uninstalled from an implementation, I'd still expect a module to start up properly, as long as it's metadata was already loaded?

This would cause some malfunctioning already(as is the case) since Iniz does load the i18n message properties necessary for proper cache initialization from the now proxy `addresshierarchy` domain.

api/src/main/java/org/openmrs/module/initializer/Domain.java
mogoodrich: Agree, I don't think it makes much of a different (yet?), but loading locations makes the most sense to me.

readme/addresshierarchy.md
```diff
@@ -0,0 +1,55 @@
...
+  <nameMapping>Province</nameMapping>
+  <sizeMapping>40</sizeMapping>
+  <requiredInHierarchy>true</requiredInHierarchy>
+</addressComponent>
+```
+
+**Option 2: if  the address level entries CSV file doesn't have a header row**, it maps the address template fields <u>in the order in which they are found in the address template</u>.
+
+###### I18n `.properties` files *(optional)*
+* The i18n `.properties` files are used to populate the address/location i18n cache. The messages files can be similarly defined as those loaded by the [messageproperties](readme/messageproperties.md) domain.
```
mks-d, line 46: I'm assuming this is optional, correct? So this is not just for Ext I18n reverse cache? Could you expand further, I don't think it's clear to the reader what this is about and how it should be used.

mks-d (changes requested):

mks-d (approved):

Issue #255, opened 2023-11-04
Liquibase loading fails when trying to execute sqlFiles relative to changset


mseaton:
Due to Liquibase upgrades, when using the existing Liquibase domain in OpenMRS versions 2.5+, relative files referenced in changesets are not able to be found.

This was fixed in core in https://issues.openmrs.org/browse/TRUNK-6077, and requires some changes that are specific to this version to work in OpenMRS 2.5+.  Specifically, liquibase files passed in need to be paths that are relative to the OpenMRS application data directory.

mseaton:
@mks-d / @ibacher / @Ruhanga / @rbuisson - see PR here:  https://github.com/mekomsolutions/openmrs-module-initializer/pull/256

mks-d:
Thanks a bunch @mseaton 💪 🙏

PR #256, opened 2023-11-04
[#255] Fixing 'liquibase' domain issues against Core 2.5.5+.

mseaton:
The genesis of this is trying to use the liquibase domain on an OpenMRS 2.5.x instance to load a liquibase file that contains "sqlFile" changesets which refer to sql files at a relative path, which fails due to an NPE.

This is something I had noticed a long time ago, and we worked around in our main PIH EMR distribution by a) patching core, and b) using a custom liquibase loader at startup.  You can [see that code here](https://github.com/PIH/openmrs-module-pihcore/blob/master/api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java), which has been used successfully in that distribution for years.  This also addresses an issue that I brought up on [the core ticket](https://issues.openmrs.org/browse/TRUNK-6077) which provided the necessary fixes in 2.5.x and beyond, where we discovered an issue with the previous use of absolute file paths in the liquibase changelog.  Basically, the use of absolute paths made the changesets non-portable, so migrating (say) from `/home/tomcat7/.OpenMRS` to `/home/tomcat8/.OpenMRS` and then restarting your sever would result in all of the changes in the liquibase file getting re-executed, whether they already had been or not, since the change in file would be interpreted by liquibase as a totally new set of changesets.  This could have very bad consequences if, for example, there are data migration or data cleanup changesets that were intended to run only once and never again and don't have sufficient preconditions on them.

This PR aims to solve these issues for implementations that have reached OpenMRS 2.5 (although this will only work for OpenMRS 2.5.5+ due to where the core patch was installed).

@mks-d / @Ruhanga / @ibacher  / @rbuisson let me know what you think.


pom.xml
```diff
@@ -54,6 +55,7 @@
         <openmrsVersion2.2>2.2.0</openmrsVersion2.2>
         <openmrsVersion2.3>2.3.6</openmrsVersion2.3>
         <openmrsVersion2.4>2.4.0</openmrsVersion2.4>
+        <openmrsVersion2.5>2.5.6</openmrsVersion2.5>
```
ibacher, line 58: Is 2.5.6 required? If so, we should probably also update that in the `@OpenmrsProfile` for the `LiquibaseLoader` (in case we develop the need for other 2.5-minimum metadata functionality).

ibacher (approved):
This seems like a good idea to me!

pom.xml
mseaton: It's actually 2.5.5. I had changed to 2.5.6 while struggling to fix some failing unit tests, but wasn't actually related to this. I've changed it back to 2.5.5 and also updated the LiquibaseLoader version requirements to match at your recommendation

ibacher:
@mseaton So, I'm running into issues with this change. Specifically, the `relativeFilePath` doesn't seem to work at all. The code in the PIH EMR seems to work because [here](https://github.com/PIH/openmrs-module-pihcore/blob/4db4a0bbaf7e469b5b75bc11f8857ef2c0400694/api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java#L28) you're still passing an absolute path, but [in this PR](https://github.com/mekomsolutions/openmrs-module-initializer/pull/256/files#diff-929a78fcf608e7869d67a826e33b94e350b55d216dc317872797c76a370f5458R40), we're only passing a relative path and this doesn't seem to work.

mseaton:
> because [here](https://github.com/PIH/openmrs-module-pihcore/blob/4db4a0bbaf7e469b5b75bc11f8857ef2c0400694/api/src/main/java/org/openmrs/module/pihcore/setup/LiquibaseSetup.java#L28) you're still passing an absolute path

That isn't an absolute path @ibacher .  It's a path relative to the application data directory, specifically "configuration/pih/liquibase/xxx.xml"

ibacher:
@mseaton Yeah, I discovered a different bug in core that needs fixing.

Issue #257, opened 2023-12-06
Concepts domain to support updating concept datatypes to numeric.


Ruhanga:
Currently it is not possible to update concepts that are worthy updating to the `Numeric` datatype. Say for example, changing existing `Text` to `Numeric`.

With this Issue resolved, it should be possible to convert concepts such as are of the `Text` datatype to `Numeric` datatype.

Related to [HSC-190](https://mekomsolutions.atlassian.net/browse/HSC-190)

[HSC-190]: https://mekomsolutions.atlassian.net/browse/HSC-190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mks-d:
@Ruhanga can this issue be closed?

Ruhanga:
Yes @mks-d. Closing it.

PR #258, opened 2023-12-07
[#257] Provision to allow overriding concepts' datatype to 'Numeric'.

Ruhanga:
Issue: #257


Ruhanga:
I agree especially if the concept was already in use by some Obs. There are those cases however where concepts can have their datatypes changed. We have [this check here](https://github.com/openmrs/openmrs-core/blob/31e9446301cc7e7d1d6a835f3d4eba0de9943b56/api/src/main/java/org/openmrs/api/impl/ConceptServiceImpl.java#L123) that enforces validity to change a concept's datatype.

mseaton (approved):
This looks right to me.  Good catch

mks-d (approved):
Approving based on others' approvals :-)

PR #260, opened 2024-01-16
O3-2446: 'queues' domain to support loading Status and Priority entries.

cioan:
@mseaton , I just added the ability to load the new queue.status_concept_set field that you added as part of the O3-2446 ticket. Thanks!

### DIFF
--- a/api-2.3/src/main/java/org/openmrs/module/initializer/api/queues/QueueLineProcessor.java
+++ b/api-2.3/src/main/java/org/openmrs/module/initializer/api/queues/QueueLineProcessor.java
@@ -19,6 +19,8 @@ public class QueueLineProcessor extends BaseLineProcessor<Queue> {
 	
 	protected static String HEADER_SERVICE = "service";
 	
+	protected static String HEADER_STATUS_CONCEPT_SET = "statusConceptSet";
+	
 	protected static String HEADER_LOCATION = "location";
 	
 	private final ConceptService conceptService;
@@ -43,6 +45,14 @@ public class QueueLineProcessor extends BaseLineProcessor<Queue> {
 				queue.setService(Utils.fetchConcept(service, conceptService));
 			} else {
 				queue.setService(null);
+			}
+		}
+		if (line.containsHeader(HEADER_STATUS_CONCEPT_SET)) {
+			String statusConceptSet = line.getString(HEADER_STATUS_CONCEPT_SET);
+			if (StringUtils.isNotBlank(statusConceptSet)) {
+				queue.setStatusConceptSet(Utils.fetchConcept(statusConceptSet, conceptService));
+			} else {
+				queue.setStatusConceptSet(null);
 			}
 		}
 		if (line.containsHeader(HEADER_LOCATION)) {
--- a/api-2.3/src/main/java/org/openmrs/module/initializer/api/queues/QueueLineProcessor.java
+++ b/api-2.3/src/main/java/org/openmrs/module/initializer/api/queues/QueueLineProcessor.java
@@ -19,7 +19,9 @@ public class QueueLineProcessor extends BaseLineProcessor<Queue> {
 	
 	protected static String HEADER_SERVICE = "service";
 	
-	protected static String HEADER_STATUS_CONCEPT_SET = "statusConceptSet";
+	protected static String HEADER_STATUS_CONCEPT_SET = "status concept set";
+	
+	protected static String HEADER_PRIORITY_CONCEPT_SET = "priority concept set";
 	
 	protected static String HEADER_LOCATION = "location";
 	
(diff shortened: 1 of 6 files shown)


mseaton:
@cioan - the issue is that this will likely fail for anyone not running the latest snapshot of queue.  It assumes that anyone loading in from the queue domain is running the latest version of the module, which is likely not true.

cioan:
Thanks @mseaton ! I just updated the dependencies to depend on the released version of the queue module.

ibacher:
> I don't see any clear existing examples of this at first glance.

I think we've been trying not to have the endless series of configurations for various versions of things. If we wanted that though, it would be possible to do two LineProcessors like this:

```java
@OpenmrsProfile(modules = { "queue:* - 1.*" })
```

```java
@OpenmrsProfile(modules = { "queue:2.*" })
```

api-2.3/src/main/java/org/openmrs/module/initializer/api/queues/QueueLineProcessor.java
```diff
@@ -19,6 +19,8 @@ public class QueueLineProcessor extends BaseLineProcessor<Queue> {
 	
 	protected static String HEADER_SERVICE = "service";
 	
+	protected static String HEADER_STATUS_CONCEPT_SET = "statusConceptSet";
```
mseaton, line 22: This does not really match the existing standard in initializer.  Have a look at some other LineProcessor classes, the headers tend to be more human readable (spaces between words, not camel case, etc).

api-2.3/src/main/java/org/openmrs/module/initializer/api/queues/QueueLineProcessor.java
```diff
@@ -45,6 +47,14 @@ public Queue fill(Queue queue, CsvLine line) throws IllegalArgumentException {
 				queue.setService(null);
 			}
 		}
+		if (line.containsHeader(HEADER_STATUS_CONCEPT_SET)) {
+			String statusConceptSet = line.getString(HEADER_STATUS_CONCEPT_SET);
+			if (StringUtils.isNotBlank(statusConceptSet)) {
```
mseaton, line 52: The key thing here is to ensure this is backwards compatible.  I _think_ this should be fine - the `setStatusConceptSet` method is not invoked unless the appropriate header is present, so for existing implementations that have an earlier version without this method and also CSV files that don't contain this header, I think this should be ok.  It would be worth verifying/testing though.

api-2.3/src/main/java/org/openmrs/module/initializer/api/queues/QueueLineProcessor.java
```diff
@@ -19,6 +19,8 @@ public class QueueLineProcessor extends BaseLineProcessor<Queue> {
 	
 	protected static String HEADER_SERVICE = "service";
 	
+	protected static String HEADER_STATUS_CONCEPT_SET = "statusConceptSet";
+
```
mseaton, line 23: Whatever code you have added to handle `statusConceptSet`, you also need to add in to handle `priorityConceptSet`, which was added to Queue at the same time.  We should cover both of these in this update.

mseaton (changes requested):
See my comments in the PR @cioan .  A few other things:
* You will need to update the README as part of this PR to reflect these additions.  Make sure you indicate that these are supported for Queue version 2.1.0 and higher.
* Will will need to update the release notes as a part of this PR.

cioan:
Thanks @mseaton ! I have addressed all your comments above. I also add @ibacher and @mks-d to the reviewers list. Thanks!

mseaton:
> I have addressed all your comments above

@cioan - please see comment above about adding to release notes and readme.  Also, I didn't see @ibacher added as a reviewer, so I added him.

mseaton:
> I think we've been trying not to have the endless series of configurations for various versions of things

In this case, I don't think we need to do anything, as the lines that hit these methods are only executed if the headers are configured for them.  Would that match your understanding @cioan and @ibacher ?

ibacher:
@mseaton Yes, I think this should work as long as the documentation is clear. @cioan We need to update [this README file](https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/queues.md#domain-queues) with the new options.

Looking at things, it looks like the queue support is only part of the 2.7.0-SNAPSHOT? In that case, it's much easier to just say that the queues domain requires version 2.1.0. This should have a very minimal effect on anyone using the existing domain, as that's likely no one currently.

cioan:
Thanks @mseaton and @ibacher ! I have updated the queues readme file. 
@mseaton , where is the release notes file that needs to be updated?

ibacher:
Usually, it's in the main README. In this case, since we haven't done a release since we've added the queue module, I think it's already covered.

mseaton:
> Usually, it's in the main README. In this case, since we haven't done a release since we've added the queue module, I think it's already covered

Agreed.  @cioan no need to update the release notes, since support for queues has actually not been in a release yet, and what we put in for 2.7.0 covers queues generally.

ibacher (approved):

ibacher:
@mseaton If you're happy with this, do you mind approving? I think this is ready to merge.

mseaton (approved):

cioan:
Thanks @mseaton and @ibacher !

Issue #263, opened 2024-03-06
OCL loader should throw an exception upon failure


mseaton:
The current implementation ocl domain simply retrieves an instance of the OCL importer, and executes the `importer.run(zip)` method on the zip files contained within the configuration domain folder.

However, when this method fails, it does not throw an exception, but rather saves all of the details of the import including any errors, if any, in the database within an Import record.  As a result, the current implementation of the OCL Loader does not throw any exceptions if the OCL loading fails.

The goal of this ticket is to fix this behavior and to ensure the OCL domain throws an Exception if loading of a zip file fails.

@mogoodrich / @ibacher / @mks-d FYI

mogoodrich:
Wow, yep, crucial, especially because we have downstream concepts that need to be installed after it.  Not sure how installing from scratch worked for us in testing with this behaviour.   @mks-d  this will be critical for us to fix.

mseaton:
I wondered the same thing @mogoodrich , and I could be wrong about this actually.  This may be a non-issue, looking back at it again now...

mseaton:
I have issued a PR here for this:

https://github.com/mekomsolutions/openmrs-module-initializer/pull/264

@mogoodrich / @ibacher / @rbuisson thoughts?

mseaton:
> We probably need to clean-up the metadata in the RefApp before merging a fix for this in, though

That may not be necessary @ibacher .  I wonder if you could test that locally with this PR applied.  From what I recall with Iniz, it is (a bit to my chagrin) configured to log errors and continue processing by default.  So for implementations that use Iniz and don't know any better, if metadata fails to load on startup the system will still continue to start as if nothing had gone wrong, and errors will only appear down the road during usage if something is amiss.

A while back, we added a configuration setting to allow implementations to change this behavior, and we have this behavior set to throw exceptions and fail hard at that time.  See https://github.com/mekomsolutions/openmrs-module-initializer/blob/master/readme/rtprops.md#4-initializerstartupload-optional

mseaton:
So as a follow-up @ibacher , I was installing a local version of the refapp to test out an unrelated ticket, and I ran the install with the latest version of Iniz that includes these changes, and it behaves as expected - there are errors in the logs due to one currently failing OCL zip, but all of the other OCL zips process subsequently, as do the other domains, and startup proceeds to completion.  Personally, I think this is a mistake for most implementations, but this PR should not adversely impact the refapp.

PR #264, opened 2024-03-06
#263 OCL loader should throw an exception upon failure

mseaton:


mogoodrich (approved):
I haven't looked at the underlying importer code, but assuming it works the way I expect, makes sense...

mseaton:
@ibacher / @rbuisson this is something we would like to get merged pretty soon.  Any feedback?

api/src/main/java/org/openmrs/module/initializer/api/loaders/OpenConceptLabLoader.java
```diff
@@ -25,7 +28,29 @@ protected String getFileExtension() {
 	protected void load(File file) throws Exception {
 		ZipFile zip = new ZipFile(file);
 		Importer importer = Context.getRegisteredComponent("openconceptlab.importer", Importer.class);
+		ImportService importService = Context.getService(ImportService.class);
+		
+		Import lastImport = importService.getLastImport();
+		log.debug("Starting OCL importer");
 		importer.run(zip);
```
mseaton, line 35: This will never throw an exception back here, so we need to inspect the ocl import table that OCL uses to report information, including errors, about the executed jobs.  I've seen a few different scenarios:

1. OCL fails to start an import at all, and so there is no ocl import entry written to the DB.  This happens if, say, the server stopped during a previous import, and the previous import row was left with a null completion date.  This will result in OCL logging stack traces to the console, but no row is written to the import table at all.  So if the new import never started (no new row in the DB), we need to throw an exception here.

2. OCL starts, and reports an error in the import table.  If we have a specific error, throw and exception with those details

3. OCL doesn't complete successfully and no specific error is logged.  Not sure when or if this would happen, but it would be an error case.

mseaton:
@ibacher / @rbuisson / @mks-d - I have updated my PR here.  Please review.

rbuisson (approved):
Very much looks like this is needed. Approved.

Issue #265, opened 2024-05-20
Spring context to shutdown before subsequent Validator invocations.


Ruhanga:
In the context of the [OpenMRS Packager plugin](https://github.com/openmrs/openmrs-contrib-packager-maven-plugin), when the Validator is used more than once during the same Maven build, the current Spring context is carried over and reused during subsequent validations (given there are more than one submodules requiring validation.)

Since some runtime properties may need to change across invocations (such as database connection properties for example), it is critical that the Spring context is properly shutdown carried prior to subsequent Validator invocations.

mks-d:
@Ruhanga this is fixed right?

PR #266, opened 2024-05-20
[#265]: Validator to properly shutdown, so not to fail further invocations on the same JVM process.

Ruhanga:
#265

### DIFF
--- a/validator/src/main/java/org/openmrs/module/initializer/validator/ConfigurationTester.java
+++ b/validator/src/main/java/org/openmrs/module/initializer/validator/ConfigurationTester.java
@@ -24,6 +24,9 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Stream;
+
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.annotation.Rollback;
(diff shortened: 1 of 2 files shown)


validator/src/main/java/org/openmrs/module/initializer/validator/ConfigurationTester.java
```diff
@@ -53,12 +56,15 @@
 import org.testcontainers.containers.MySQLContainer;
 import ch.vorburger.exec.ManagedProcessException;
 
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
```
ibacher, line 59: I'm not sure if this annotation does anything because of how `BaseContextSenstitiveTest` is defined.

ibacher (approved):
It's probably safest to add a `null` check before calling close (so we don't accidentally hide a real issue), but this seems fine.

validator/src/main/java/org/openmrs/module/initializer/validator/ConfigurationTester.java
Ruhanga: It actually does, as it forces the recreation of a new Spring application context on the next run. Also `BaseContextSenstitiveTest` extends `AbstractJUnit4SpringContextTests` which theoretically means being able to hook into the Spring context life-cycle.

PR #273, opened 2024-07-08
Added support to reload the Ampath Forms translations 

icrc-loliveira:
Added support to reload the Ampath Forms translations every time a form is changed.

- Added checksums files to  Ampath Forms translations.
- Remove the check sums for the Ampath Forms translations when a new Ampath Forms is loaded if already exist translation for an older version.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsLoader.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsLoader.java
@@ -1,7 +1,9 @@
 package org.openmrs.module.initializer.api.loaders;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.io.FileUtils;
@@ -15,6 +17,7 @@ import org.openmrs.api.EncounterService;
 import org.openmrs.api.FormService;
 import org.openmrs.api.db.ClobDatatypeStorage;
 import org.openmrs.module.initializer.Domain;
+import org.openmrs.module.initializer.api.ConfigDirUtil;
 import org.openmrs.module.initializer.api.utils.Utils;
 import org.openmrs.util.OpenmrsUtil;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -25,6 +28,8 @@ import java.util.UUID;
 public class AmpathFormsLoader extends BaseFileLoader {
 	
 	public static final String AMPATH_FORMS_UUID = "794c4598-ab82-47ca-8d18-483a8abe6f4f";
+	
+	public static final String JSON_EXTENSION = "json";
 	
 	@Autowired
 	private DatatypeService datatypeService;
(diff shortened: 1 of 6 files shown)


icrc-loliveira:
Hello @ibacher,
Can you please review this PBI? 

cc: @icrc-fdeniger

api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsTranslationsLoader.java
```diff
@@ -59,20 +59,16 @@ protected void load(File file) throws Exception {
...
-			if (formName == null) {
-				throw new IllegalArgumentException("'form' property is required for AMPATH forms translations loader.");
-			}
-			form = formService.getForm(formName);
-			if (form == null) {
-				throw new IllegalArgumentException(
-				        "Could not find a form named '" + formName + "'. Please ensure an existing form is configured.");
-			}
+			throw new IllegalArgumentException(
+			        "Could not find a form named '" + formName + "'. Please ensure an existing form is configured.");
```
ibacher, line 71: Could you explain why you did this re-write?

ibacher: Which also seems more correct, because otherwise we end up with historical forms for which we have no translation strings.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsLoader.java
```diff
@@ -83,6 +88,25 @@ protected void load(File file) throws Exception {
 			throw new Exception("Form Version is required");
 		}
 		
+		// Delete Checksum Files for the translation files associated with the form
+		ConfigDirUtil configDirUtil = new ConfigDirUtil(iniz.getConfigDirPath(), iniz.getChecksumsDirPath(),
+		        Domain.AMPATH_FORMS_TRANSLATIONS.getName(), true);
+		
+		List<File> translationFiles = configDirUtil.getFiles(JSON_EXTENSION);
```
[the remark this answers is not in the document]
icrc-loliveira: Done, I was just not able to keep the line bellow because 1 form may have multiple languages resources.
```break;  // once we've found the form, we're done.```

api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsTranslationsLoader.java
icrc-loliveira: I have done this because in other way it would always update the existing form.

icrc-loliveira: I have done your suggestion, and because know we are not able to use the `uuid`, in order to get an existing resource in case it needs to be updated, and not duplicated, it will check that the resource have the key `translations` and that the value of the key `language` is the same.

icrc-loliveira:
Hello @ian,
Thanks for your review I have done the requested changes.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsLoader.java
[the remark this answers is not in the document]
ibacher: Ah! That's a good catch!

ibacher (approved):

ibacher:
Alright. Code is fine, but we need some documentation, specifically two things:

1. In the docs for the ampathformloader, we should explain that any translations will be reloaded when a form is updated or added.
2. In the main README, we should add a line to the release notes for the 2.8.0 version about this change.

Those two things, and I think this can be merged.

icrc-loliveira:
Hello @ibacher,
I have updated the readme files as requested.

ibacher (approved):

PR #275, opened 2024-08-09
Docs: Update ampathformstranslations.md

gracepotma:
It was not yet clear to me until I learned the hard way that you need an English translation file even if the original JSON form schema was in english.

### DIFF
--- a/readme/ampathformstranslations.md
+++ b/readme/ampathformstranslations.md
@@ -7,6 +7,8 @@ ampathformstranslations/
   └── form2_translations_en.json
 ```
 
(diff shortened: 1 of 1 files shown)


readme/ampathformstranslations.md
```diff
@@ -7,6 +7,8 @@ ampathformstranslations/
   └── form2_translations_en.json
 ```
 
+**Note:** You will need a translation file for ALL languages you want to be able to switch between. For example, if you want to be able to switch between both French and English, you would need a file for BOTH those languages, as shown above. If you only had a fr.json file, then even when you set your language preference to English, you would still be seeing that form only in French. 
+
```
mks-d, line 11: >You will need a translation file for ALL languages you want to be able to switch between.

@gracepotma isn't this obvious tho? Or am I missing the point?

gracepotma: It wasn't obvious to me :P

Issue #284, opened 2024-10-24
Support "dispositions" domain


mogoodrich:
The EMR API module expects a "dispositionConfig.json" file to be present in order to configure it's disposition handling logic.

Currently, the two main ways to set up this file are installing it on the classpath or setting it up programmatically.

We'd like to set this up via Iniz... specifically, if there is a "*.json" file in the "configuration/dispositions" directory, this should be loaded as the dispositions configuration file.

mogoodrich:
PR here: https://github.com/mekomsolutions/openmrs-module-initializer/pull/285

PR #285, opened 2024-10-24
[#284] - Support "dispositions" domain

mogoodrich:
https://github.com/mekomsolutions/openmrs-module-initializer/issues/284


pom.xml
```diff
@@ -66,6 +66,7 @@
         <bahmniIeAppsVersion>1.1.0</bahmniIeAppsVersion>
         <bahmniCoreVersion>0.93-1.1.0</bahmniCoreVersion>
         <cohortModuleVersion>3.5.0</cohortModuleVersion>
+        <emrapiModuleVersion>2.1.0-SNAPSHOT</emrapiModuleVersion>
```
mogoodrich, line 69: Right now we need a change I just made in EMR-API... once that is approved, I can merge and release 2.1.0 and update here before merging this in.

mseaton: Do we?  See my comment above...

api/src/main/java/org/openmrs/module/initializer/api/loaders/DispositionsLoader.java
```diff
@@ -0,0 +1,50 @@
...
+	}
+	
+	@Override
+	protected void load(File file) throws Exception {
+		if (fileFound) {
+			throw new IllegalArgumentException(
+			        "Multiple disposition files found in the disposition configuration directory.");
+		}
+		fileFound = true;
+		dispositionService.setDispositionConfig("file:" + file.getAbsolutePath());
```
mseaton, line 42: We should be able to relativize this path to the application data directory using `Path.relativize`, and then set that the dispositionConfig as expected.  We have made some efforts in the past due to security issues to not allow passing in files outside of the application data directory, so I think it's fine to expect/require any file to be relative to this location.

Then, this does not need to depend on the latest snapshot of emrapi, it can support versions going way back.

mogoodrich: this makes sense, working on this now.

api/src/main/java/org/openmrs/module/initializer/api/InitializerService.java
```diff
@@ -20,6 +21,11 @@
 
 public interface InitializerService extends OpenmrsService {
 	
+	/**
+	 * @return A Path object representing the base path of the application data directory
+	 */
+	Path getBasePath();
```
mogoodrich, line 27: This is a public method in the InitializerServiceImpl that I wanted to use, but it wasn't on the interface for some reason, so I added it... I can certainly go about this another way if there is a reason it wasn't exposed on the interface.

api/src/main/java/org/openmrs/module/initializer/api/loaders/DispositionsLoader.java
```diff
@@ -0,0 +1,50 @@
...
+	}
+	
+	@Override
+	protected void load(File file) throws Exception {
+		if (fileFound) {
+			throw new IllegalArgumentException(
+			        "Multiple disposition files found in the disposition configuration directory.");
+		}
+		fileFound = true;
+		dispositionService.setDispositionConfig("file:" + iniz.getBasePath().relativize(file.toPath()));
```
mogoodrich, line 42: LMK if this looks like the right way to do this.

pom.xml
```diff
@@ -66,6 +66,7 @@
         <bahmniIeAppsVersion>1.1.0</bahmniIeAppsVersion>
         <bahmniCoreVersion>0.93-1.1.0</bahmniCoreVersion>
         <cohortModuleVersion>3.5.0</cohortModuleVersion>
+        <emrapiModuleVersion>2.0.0</emrapiModuleVersion>
```
mogoodrich, line 69: We could likely go lower than this if we want, but tests were failing with versions lower than this with autowire issue.  Assumedly could fix by including emrapi-api-x.x.x submodules that were removed in EMR API 2.0.0, but didn't seem worth it.  We want to encourage people to update to 2.0.0.

api/src/main/java/org/openmrs/module/initializer/api/loaders/DispositionsLoader.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.emrapi.disposition.DispositionService;
+import org.openmrs.module.initializer.Domain;
+import org.openmrs.module.initializer.api.ConfigDirUtil;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.io.File;
+
+@OpenmrsProfile(modules = { "emrapi:2.1.0-SNAPSHOT-9.*" })
```
mseaton, line 11: Per the version you set in the pom, this should be set accordingly to 2.0.0-9.*

api/src/main/java/org/openmrs/module/initializer/api/loaders/DispositionsLoader.java
mseaton: Looks right.

api/src/test/java/org/openmrs/module/initializer/DomainBaseModuleContextSensitiveTest.java
```diff
@@ -131,6 +131,11 @@ public DomainBaseModuleContextSensitiveTest() {
 			mod.setFile(new File(""));
 			ModuleFactory.getStartedModulesMap().put(mod.getModuleId(), mod);
 		}
+		{
+			Module mod = new Module("", "emrapi", "", "", "", "2.1.0-SNAPSHOT");
```
mseaton, line 135: Per the version you set in the pom, this should be set accordingly to 2.0.0

readme/dispositions.md
```diff
@@ -0,0 +1,14 @@
...
+**Dispositions** subfolder contains a single JSON that defines the dispositions available in the system::
+
+```bash
+dispositions/
+  ├──dispositions.json
+```
+
+#### Requirements
+* The [emr-api](https://github.com/openmrs/openmrs-module-emrapi) version 2.0.0 or higher must be installed
+* The OpenMRS version must be 2.2.1 or higher most be installed
```
mseaton, line 11: Why?  Is because emrapi 2.0.0 requires OpenMRS 2.2.1?  I wouldn't mention that, as that becomes a rabbit hole of dependency versions that we need to alert about for every top-level dependency.

api/src/test/java/org/openmrs/module/initializer/api/DispositionsLoaderIntegrationTest.java
```diff
@@ -0,0 +1,28 @@
...
+	private DispositionsLoader loader;
+	
+	@Autowired
+	private DispositionService dispositionService;
+	
+	@Test
+	public void load_shouldLoadDisposition() {
+		loader.load();
+		assertEquals(5, dispositionService.getDispositions().size());
+	}
```
mseaton, line 26: Seems like we should have 1 or 2 other tests.  eg:
* Test that dispositions are always reloaded every time
* Test that an exception is thrown if multiple files are found in the domain

api/src/main/java/org/openmrs/module/initializer/api/loaders/DispositionsLoader.java
mogoodrich: oh, good point, will fix

api/src/test/java/org/openmrs/module/initializer/DomainBaseModuleContextSensitiveTest.java
mogoodrich: Again, good point, will fix

readme/dispositions.md
mogoodrich: yeah, I just copied this file from another example domain and it listed both the module dependency and core dependency, and, yes, EMR API 2.0.0 depends on 2.2.1 core, but you are right, not really necessary to include it here.

api/src/test/java/org/openmrs/module/initializer/api/DispositionsLoaderIntegrationTest.java
mogoodrich: New tests added, see above.

It's slightly hacky that the duplicate config file is present for all three of the tests above so we can test the "exception" case, and we are relying on the fact that Iniz ignores exceptions by default test the happy paths.

mogoodrich: I also confirmed that the reload test does fail if we don't have checksums properly disabled.

mogoodrich:
Updated based on Mike's feedback, ready for re-review, pending results of build and test run.

mseaton (approved):

api/src/main/java/org/openmrs/module/initializer/api/InitializerService.java
Ruhanga: The function `getBasePath()` was introduced to fix a bug which I guess should have been kept private, but it seems to broadly expose the application directory with content that falls outside the designated folder for a given domain. The design concern I'd have is more to do with encapsulation here, particularly in relation to Iniz Loaders. Happy to have the change in though following @mseaton's approval.

Ruhanga (commented):
Generally looking good. I have a few minor concerns in comment below...

api/src/test/java/org/openmrs/module/initializer/api/DispositionsLoaderIntegrationTest.java
mogoodrich: Oh,nice, thanks @Ruhanga , I can try that.

api/src/main/java/org/openmrs/module/initializer/api/InitializerService.java
mogoodrich: Not sure I follow @Ruhanga  but basically I need access to the base path to do the "relatizing" on DispositionsLoader:42... if there's a better way to do this (or if this isn't a "safe" way to fetch the base path) happy to change it.  @mseaton

mseaton: Well, we already have core methods in OpenmrsUtil to get the application data directory as a String or as a File, so if all this is doing is exposing the same as a path, that seems perfectly safe to me.

Ruhanga: Got it @mseaton! I'm happy to have this in if everything looks good to you. Thanks.

Ruhanga (commented):
Thanks @mogoodrich. I believe this should work with the validator too, and so the relevant emrapi packages should be packaged with it. Is this something you've tried out already?

mogoodrich:
> Thanks @mogoodrich. I believe this should work with the validator too, and so the relevant emrapi packages should be packaged with it. Is this something you've tried out already?

I don't know what your are talking about @Ruhanga , so probably not... :)  Is there a Maven profile or something I need to trigger to run some sort of validation tests?

mogoodrich:
> I don't know what your are talking about @Ruhanga , so probably not... :) Is there a Maven profile or something I need to trigger to run some sort of validation tests?

Actually, @mseaton pointed out to me that the validator tests run on the PR and passed, so I think we are good?

mseaton:
I think it looks ready to merge.  @Ruhanga ?

Ruhanga:
> Actually, @mseaton pointed out to me that the validator tests run on the PR and passed, so I think we are good?

I apologize for the confusion. There isn’t much that can simulate the validator in unit tests. What I was referring to was conducting a runtime test of the validator using a minimal configuration related to the `dispositions` domain, as outlined in [the documentation](https://github.com/mekomsolutions/openmrs-module-initializer/blob/main/readme/validator.md#dry-run-examples).

mogoodrich:
> I apologize for the confusion @mseaton, @mogoodrich. There isn’t much that can simulate the validator in unit tests. What I was referring to was conducting a runtime test of the validator using a minimal configuration involving the `dispositions` domain, as outlined in [the documentation](https://github.com/mekomsolutions/openmrs-module-initializer/blob/main/readme/validator.md#dry-run-examples).

Ah, got it @Ruhanga ... I confirmed that running the validator against the dispositions domain was failing because the validator pom.xml was missing an emrapi dependency.  I added it and confirmed that things work.

validator/pom.xml
```diff
@@ -398,6 +398,14 @@
...
     </dependency>
 
+    <dependency>
+      <groupId>org.openmrs.module</groupId>
+      <artifactId>emrapi-api</artifactId>
+      <version>${emrapiModuleVersion}</version>
+      <scope>runtime</scope>
+      <type>jar</type>
+    </dependency>
+
```
mogoodrich, line 408: Added EMR API dependency to validator domain so it can properly validate disposition loading.

README.md
```diff
@@ -159,6 +160,7 @@ This is the list of currently supported domains in their loading order:
 1. [AMPATH Forms (JSON files)](readme/ampathforms.md)
 1. [AMPATH Forms Translations (JSON files)](readme/ampathformstranslations.md)
 1. [HTML Forms (XML files)](readme/htmlforms.md)
+1. [Disposition Config (json files)](readme/dispositions.md)
```
Ruhanga, line 163:
```suggestion
1. [Disposition Config (json file)](readme/dispositions.md)
```

Ruhanga (approved):
Generally LGTM, save a doc correction to make before the merging I guess. Thanks @mogoodrich, approving it for now.

Issue #288, opened 2025-01-30
Global Properties Config Serializer fails on OpenMRS 2.7+


mseaton:
Due to the version of XStream used in OpenMRS Core 2.7+, additional configuration needs to be done on new instances of XStream serializers to allow deserialization of specific classes.  Without such a change, trying to load in global property xml files in an OpenMRS 2.7.x environment results in a failure, with an error like the following:

```com.thoughtworks.xstream.security.ForbiddenClassException: org.openmrs.module.initializer.api.gp.GlobalPropertiesConfig```

mseaton:
PR here:  https://github.com/mekomsolutions/openmrs-module-initializer/pull/289

@ibacher / @Ruhanga - this is a blocker for using Iniz on the latest versions of OpenMRS

mseaton:
This fix has been merged (see commit above).  Closing this out.  Thanks @Ruhanga !

PR #289, opened 2025-01-30
#288 - Fix Global Properties loading for OpenMRS 2.7+

mseaton:


api/src/main/java/org/openmrs/module/initializer/api/InitializerSerializer.java
```diff
@@ -53,6 +57,15 @@ public static XStream getGlobalPropertiesConfigSerializer() {
...
 		xs.aliasField("value", GlobalProperty.class, "propertyValue");
+		try {
+			Method allowTypeHierarchy = XStream.class.getMethod("allowTypeHierarchy", Class.class);
+			allowTypeHierarchy.invoke(xs, GlobalPropertiesConfig.class);
+			allowTypeHierarchy.invoke(xs, GlobalProperty.class);
+			log.debug("Successfully configured global properties config serializer with allowed types");
+		}
+		catch (Exception e) {
+			log.debug("Error configuring global properties config serializer with allowed types", e);
+		}
```
Ruhanga, line 68: This logic seems to imply that it fails on older versions of core. We might not necessarily need a separate `InitializerSerializer_2_7`, but it could be useful to have some unit test(s) enforcing the logic .

mseaton: Not sure what you mean about `this logic seems to imply that it fails on older versions of core`.  I don't think it implies that.  But your point about having some unit tests that actually show this failure and prove that this fixes it is totally valid.  I copied the test from the `api` project (which tests using OpenMRS 2.1) into the `api-2.7` project (which tests using OpenMRS 2.7).  This test fails without my fix above, which is easily testable if you comment out lines 62 and 63 above.

Ruhanga: > Not sure what you mean about `this logic seems to imply that it fails on older versions of core`.

@mseaton, sorry for the confusion. I was just wondering why the error is logged instead of actually re-throwing a runtime exception.

mseaton: I see @Ruhanga .  It is logged as it is assumed that any Exception thrown by this block is due to the fact that this method is not present in the Xstream library, which is an expected condition for any platform < 2.7.

Ruhanga: Alright, makes sense to me. Thanks @mseaton.

Ruhanga (approved):

Issue #290, opened 2025-01-30
DeleteDomainChecksumsChangesetIntegrationTest fails in some environments


mseaton:
Trying to build the module from latest today, I consistently ran into a test failure with the `DeleteDomainChecksumsChangesetIntegrationTest` test.

Investigating this a little bit, I believe this is due to a dirtied context that affects the conditions this test needs to execute, as this test passes on it's own and currently passes on github actions, but fails in my local environment.

My suspicion is that this class depends on some specific LIQUIBASECHANGELOG and LIQUIBASECHANGELOGLOCK tables existing, yet there are some tests that create and delete these tables within the context, namely the teardown method of `LiquibaseLoader25IntegrationTest` which runs:

```
	@After
	public void teardown() {
		adminService.executeSQL("drop table LIQUIBASECHANGELOG", false);
		adminService.executeSQL("drop table LIQUIBASECHANGELOGLOCK", false);
	}
```

mseaton:
PR here:  https://github.com/mekomsolutions/openmrs-module-initializer/pull/291
Not really the most elegant solution, and doesn't really address the underlying issue, but it does fix things.

mseaton:
Committed, see above.  Thanks @Ruhanga

PR #291, opened 2025-01-30
#290 - Fix collisions between unit tests

mseaton:


Ruhanga (approved):
I don't anticipate any issues with this. LGTM!

PR #294, opened 2025-02-18
Upgraded validator's OpenMRS core to 2.6.9.

Ruhanga:
Upgrades validator to OpenMRS Core 2.6.9.

### DIFF
--- a/api-2.4/src/test/java/org/openmrs/module/initializer/api/BillableServicesLoaderIntegrationTest.java
+++ b/api-2.4/src/test/java/org/openmrs/module/initializer/api/BillableServicesLoaderIntegrationTest.java
@@ -21,7 +21,7 @@ import org.openmrs.module.initializer.api.billing.BillableServicesLoader;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 
-public class BillableServicesLoaderIntegrationTest extends DomainBaseModuleContextSensitive_2_4_test {
+public class BillableServicesLoaderIntegrationTest extends DomainBaseModuleContextSensitive_2_4_Test {
 	
 	@Autowired
 	@Qualifier("conceptService")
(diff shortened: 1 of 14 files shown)


validator/pom.xml
```diff
@@ -405,6 +416,34 @@
...
+	  <type>jar</type>
+	</dependency>
+
+  <dependency>
+    <groupId>org.liquibase.ext</groupId>
+    <artifactId>liquibase-compat</artifactId>
+    <type>jar</type>
+    <scope>runtime</scope>
+    <version>4.0.0-beta2</version>
+  </dependency>
```
mks-d, line 446: Slight indentation issue here.

api-2.4/src/test/java/org/openmrs/module/initializer/api/BillableServicesLoaderIntegrationTest.java
```diff
@@ -21,7 +21,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 
-public class BillableServicesLoaderIntegrationTest extends DomainBaseModuleContextSensitive_2_4_test {
+public class BillableServicesLoaderIntegrationTest extends DomainBaseModuleContextSensitive_2_4_Test {
```
mks-d, line 24: Not sure what got changed here? There's a bunch of files with similar changesets... 🤔

mks-d (commented):
Thanks @Ruhanga! I'm assuming this stems from the fact that the Validator's promise is to validate against the latest available Core version? I don't remember how this was originally set up to be honest.

Ruhanga:
> I'm assuming this stems from the fact that the Validator's promise is to validate against the latest available Core version? I don't remember how this was originally set up to be honest.

@mks-d, you're correct about the Validator's promise to validate against the latest available Core version. However, this hasn't been a priority in recent times.

Ruhanga:
Hi @mks-d, could there be any reservations to having this in already?

PR #302, opened 2025-09-26
(fix): Move form name translation to preload

NethmiRodrigo:
This PR moves form name translation from the load function to the preload function to ensure that the form name translations are available after an instance restart.

### DIFF
--- a/api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsTranslationsLoader.java
+++ b/api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsTranslationsLoader.java
@@ -41,6 +41,42 @@ public class AmpathFormsTranslationsLoader extends BaseFileLoader {
 		return "json";
 	}
 	
+	private Form getForm(Map<String, Object> jsonTranslationsDefinition) {
+		Form form;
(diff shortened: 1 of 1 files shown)


ibacher (approved):

api/src/test/java/org/openmrs/module/initializer/api/form/AmpathFormsTranslationsLoaderIntegrationTest.java
```diff
@@ -158,6 +158,7 @@ public void load_shouldThrowGivenMissingFormFieldInFormTranslationsDef() throws
 	@Test
 	public void load_shouldThrowGivenMissingLanguageFieldInFormTranslationsDef() throws Exception {
 		// Setup
+		ampathFormsLoader.load();
```
ibacher, line 161: This is a bit weird to me... However, I think it should be right before the call to `ampathFormsTranslationLoader` rather than here.

NethmiRodrigo: I think its because now the form needs to get loaded in the preload. Moved it!

PR #307, opened 2025-11-27
TRUNK-6409: Initializer module should not re-run initialization on each node

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/TRUNK-6409

## Checklist: I completed these to help reviewers :)

- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [x] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`


IamMujuziMoses:
@rkorytkowski @dkayiwa @ibacher could you please review this PR, your feedback will be appreciated.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -76,20 +77,34 @@ public void started() {
 		InitializerMessageSource messageSource = getInitializerMessageSource();
 		Context.getMessageSourceService().setActiveMessageSource(messageSource);
 		
+		boolean isPrimary = Boolean.parseBoolean(getPropertyValue(PROPS_PRIMARY_STARTUP, "false"));
```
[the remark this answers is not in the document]
rkorytkowski: @IamMujuziMoses [TRUNK-6409](https://openmrs.atlassian.net/browse/TRUNK-6409) is about persisting checksums in a dedicated table in DB instead of in filesystem and then adding a check to run import only on one replica at a time. It can be achieved similar to how liquibase does it i.e. have a semaphore like flag in db that is checked before running an import and having a node that starts the import to hold it until it's done before any other node can do it. This way we don't have to set any node as primary. The drawback is that each node will be doing a check if checksums changed and comparing to DB persisted values, but that should be quick and can be further optimised if needed. I think it's simplifies deployments. You don't have to keep 2 different deployments and remember to deploy in sequence. I'll update the issue description.

rkorytkowski (changes requested):

rkorytkowski (changes requested):
Also please note that the module already persist checksums for files in the filesystem. We need to put those checksums in DB so they are available for all nodes and not lost with filesystem failure.

api/src/main/java/org/openmrs/module/initializer/api/entities/InitializerLock.java
```diff
@@ -0,0 +1,66 @@
...
+@Entity
+@Table(name = "initializer_lock")
+public class InitializerLock {
+	
+	@Id
+	@Column(name = "id")
+	private Integer id;
+	
+	@Column(name = "locked", nullable = false)
+	private Boolean locked;
```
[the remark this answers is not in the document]
rkorytkowski: Having locks with different names would mean that if someone runs initializer in their own module, it can use a different lock name and thus run initialisation regardless of the default initialization done by the core module.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -223,4 +231,111 @@ public InitializerConfig getInitializerConfig() {
...
+		for (File f : files) {
+			if (!f.isFile()) {
+				continue;
+			}
+			try (FileInputStream fis = new FileInputStream(f)) {
+				Path rel = base.relativize(f.toPath());
+				String checksum = DigestUtils.sha256Hex(fis);
+				map.put(rel.toString(), checksum);
+			} catch (Exception e) {
+				map.put(f.getName(), "ERROR");
```
rkorytkowski: Isn't the initalizer module computing the checksums elsewhere and persisting them in the filesystem? We need to remove that old behavior.

rkorytkowski (changes requested):
Thanks! Going in the good direction. Still a few tweaks needed.

rkorytkowski:
@IamMujuziMoses thanks, I'll go through your code and do testing on Monday.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -82,14 +82,38 @@ public void started() {
...
+				
+				// Check if config changed
+				if (!getInitializerService().isConfigChanged()) {
+					log.info("No config changes... skipping initializer");
+					return;
+				}
+				
+				// Run Initializer
+				log.info("OpenMRS config loading process started...");
 				getInitializerService().loadUnsafe(true, throwError);
```
rkorytkowski, line 101: My understanding is that with your changes if only a single checksum is different, then all files are being loaded again instead of just the one that changed... It's not correct. Only that single file should run. I'll fix that issue.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -223,4 +228,144 @@ public InitializerConfig getInitializerConfig() {
...
+			for (Iterator<Path> it = stream.iterator(); it.hasNext();) {
+				
+				Path path = it.next();
+				if (!Files.isRegularFile(path)) {
+					continue;
+				}
+				
+				try (InputStream is = Files.newInputStream(path)) {
+					Path rel = base.relativize(path);
+					String checksum = DigestUtils.sha256Hex(is);
```
rkorytkowski, line 346: You changed the checksum from md5Hex to sha256Hex... which is probably fine, but you don't seem to handle migration from file stored checksums to db checksums. I'll address that as well.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -82,14 +82,38 @@ public void started() {
...
+				
+				// Save new checksums
+				getInitializerService().updateChecksums();
```
rkorytkowski, line 104: We should not updateChecksums all at once... It's possible that some files get imported and checksums should be updated and some fail... I'll address that as well.

rkorytkowski:
Thanks @IamMujuziMoses! I've left a few review comments. Let me address them as I need to get it done today.

rkorytkowski:
PR with fixes: https://github.com/mekomsolutions/openmrs-module-initializer/pull/319

This one (#307) can be closed.

PR #319, opened 2026-03-10
TRUNK-6409 Initializer module should not re-run initialisation on each node

rkorytkowski:
PR with fixes for https://github.com/mekomsolutions/openmrs-module-initializer/pull/307 

I'll continue with testing against reference application in docker compose and in cluster setup.


rkorytkowski:
@IamMujuziMoses could you please review my fixes?

rkorytkowski:
@mseaton, @brandones, @wikumChamith I'd appreciate your reviews as well since you contributed here.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsLoader.java
```diff
@@ -86,21 +86,6 @@ protected void load(File file) throws Exception {
...
-			String js = FileUtils.readFileToString(translationFile, StandardCharsets.UTF_8.toString());
-			Map<String, Object> jf = new ObjectMapper().readValue(js, Map.class);
-			String translationForm = (String) jf.get("form");
-			
-			if (StringUtils.equals(translationForm, formName)) {
-				configDirUtil
-				        .deleteChecksumFile(replaceExtension(translationFile.getName(), ConfigDirUtil.CHECKSUM_FILE_EXT));
-			}
-		}
-
```
mseaton, line 103: It seems like we should not remove this block?  Certainly we can change how we interact with checksums, but it seems like the point of this block is to delete checksums to force forms to reload that this deems to have change via some transitory change.

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseFileLoader.java
```diff
@@ -95,9 +98,6 @@ public void loadUnsafe(List<String> wildcardExclusions, boolean doThrow) throws
 					        throw new RuntimeException(e);
 				        }
 			        }
-			        
-			        dirUtil.writeChecksum(file, dirUtil.getChecksumIfChanged(file)); // the config file is marked as processed
-			        log.info("The '" + getDomainName() + "' configuration file has finished loading:\n" + file.getPath());
```
mseaton, line 100: Dd you intend to lose this log.info statement?  If so, why?

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseLoader.java
```diff
@@ -31,7 +31,7 @@ public boolean isPreLoader() {
 	
 	@Override
 	public ConfigDirUtil getDirUtil() {
-		return new ConfigDirUtil(iniz.getConfigDirPath(), iniz.getChecksumsDirPath(), getDomainName(), cfg.skipChecksums());
```
mseaton, line 34: Is this idea of "skipChecksums" implemented elsewhere?

api/src/main/java/org/openmrs/module/initializer/api/loaders/DispositionsLoader.java
```diff
@@ -45,6 +45,6 @@ protected void load(File file) throws Exception {
 	@Override
 	public ConfigDirUtil getDirUtil() {
 		// skip checksums, this needs to be processed every time
-		return new ConfigDirUtil(iniz.getConfigDirPath(), iniz.getChecksumsDirPath(), getDomainName(), true);
```
mseaton, line 48: Does this retain the "skip" behavior as documented?

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -76,27 +86,94 @@ public void started() {
 		InitializerMessageSource messageSource = getInitializerMessageSource();
 		Context.getMessageSourceService().setActiveMessageSource(messageSource);
 		
-		String startupLoadingMode = getInitializerService().getInitializerConfig().getStartupLoadingMode();
+		migrateChecksumsFromFilesToDB();
```
mseaton, line 80: It would be good if we could maybe move all of the logic below that is responsible for kicking off initializer into a standalone method that could be invoked elsewhere, if there is specific logic that we want to make sure is always run together.   At least at PIH, we disable loading via the activator, as we need it to run after all other modules have started, etc

api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
```diff
@@ -28,9 +28,4 @@ protected void load(File file) throws Exception {
 			DatabaseUpdater.executeChangelog(file.getPath(), null);
 		}
 	}
-	
-	@Override
-	public ConfigDirUtil getDirUtil() {
-		return new ConfigDirUtil(iniz.getConfigDirPath(), iniz.getChecksumsDirPath(), getDomainName(), true);
```
mseaton, line 34: Generally I am worried that we are losing the notion of having checksums deleted in order to force re-load a particular file or a particular domain in whole.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -76,27 +86,94 @@ public void started() {
...
 		
+		String startupLoadingMode = getInitializerService().getInitializerConfig().getStartupLoadingMode();
 		if (PROPS_STARTUP_LOAD_DISABLED.equalsIgnoreCase(startupLoadingMode)) {
 			log.info("OpenMRS config loading process disabled at initializer startup");
 		} else {
 			boolean throwError = PROPS_STARTUP_LOAD_FAIL_ON_ERROR.equalsIgnoreCase(startupLoadingMode);
-			log.info("OpenMRS config loading process started...");
+			boolean lockAquired = false;
```
wikumChamith: typo


```suggestion
			boolean lockAcquired = false;
```

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -76,27 +86,94 @@ public void started() {
...
+		if (globalProperty == null) {
+			// Migrate file checksums, to be removed in later versions
+			Path checksumsBase = Paths.get(getInitializerService().getChecksumsDirPath());
+			Path base = Paths.get(getInitializerService().getConfigDirPath());
+			try (Stream<Path> stream = Files.walk(base)) {
+				stream.filter(Files::isRegularFile).forEach(path -> {
+					String relConfigFilePath = base.relativize(path).toString();
+					relConfigFilePath = relConfigFilePath.replaceFirst(File.separator, "!!!") //change domain separator
+							.replace(File.separator, "_")
+							.replaceFirst("!!!", File.separator); //revert domain separator
```
wikumChamith: Would this work on Windows?  `replaceFirst() `uses regex, and on Windows, `File.separator `is` \`, which is a special regex character.

api/src/main/resources/liquibase.xml
```diff
@@ -40,5 +40,40 @@
         <comment>Deleting 'concepts' domain checksums</comment>
         <customChange class="org.openmrs.module.initializer.liquibase.DeleteDomainChecksumsChangeset" domainName="concepts"></customChange>
```
wikumChamith: Do we still need this?

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -223,4 +236,153 @@ public InitializerConfig getInitializerConfig() {
...
+	 * @see org.openmrs.module.initializer.api.InitializerService#hasChecksumsChanged()
+	 */
+	@Transactional(readOnly = true)
+	@Override
+	public boolean hasChecksumsChanged() {
+		Map<String, String> db = getAllChecksums();
+		Map<String, String> fs = getFileChecksums();
+		
+		return !db.equals(fs);
+	}
```
wikumChamith, line 298: Can't we optimize this by comparing file counts and paths before computing any checksums?

We could return early if the file count differs from the DB count.

rkorytkowski: The equals method for maps is quite clever and does compare counts first, see https://docs.oracle.com/javase/8/docs/api/java/util/AbstractMap.html#equals-java.lang.Object-

api/src/main/resources/liquibase.xml
rkorytkowski: Correct, needs to be removed.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
rkorytkowski: Probably not. It can be easily done without regex. Let me fix it.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
rkorytkowski: All right! I'll do it.

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseFileLoader.java
rkorytkowski: Reverted!

api/src/main/java/org/openmrs/module/initializer/api/loaders/BaseLoader.java
rkorytkowski: No! I missed that change in @IamMujuziMoses commit. I'll revert it.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
wikumChamith: No, we still calculate the checksum values before that. I meant, can't we compare file counts and paths before computing any checksums?

```
public boolean hasChecksumsChanged() {                                                                                                             
      Map<String, String> db = getAllChecksums();
                                                                                                                                                     
      long fileCount = countConfigFiles();   // This just returns the count of config files                                                                                          
      if (db.size() != fileCount) {                                                                                                                  
          return true;
      }                                                                                                                                              
                  
      Map<String, String> fs = getFileChecksums();
      return !db.equals(fs);                                                                                                                         
  }
```

This is just a suggestion. This could be unnecessary :)

api/src/main/java/org/openmrs/module/initializer/api/loaders/AmpathFormsLoader.java
rkorytkowski: I'll revert it.

api/src/main/java/org/openmrs/module/initializer/api/loaders/LiquibaseLoader.java
rkorytkowski: I'll revert this. I'll add deleteChecksums(String domain) and deleteChecksum(Path path) to the InitializerService.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
rkorytkowski: I'll fix it.

IamMujuziMoses:
thanks for the updates, LGTM the changes make the initialization process more robust.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
```diff
@@ -20,17 +20,11 @@ protected Domain getDomain() {
...
-		return new ConfigDirUtil(iniz.getConfigDirPath(), iniz.getChecksumsDirPath(), getDomainName(), true);
-	}
-	
 	@Override
 	public void loadUnsafe(List<String> wildcardExclusions, boolean doThrow) throws Exception {
 		
 		try {
-			AddressConfigurationLoader.loadAddressConfiguration(Paths.get(iniz.getConfigDirPath()),
-			    Paths.get(iniz.getChecksumsDirPath()));
+			AddressConfigurationLoader.loadAddressConfiguration(Paths.get(iniz.getConfigDirPath()), null);
```
rkorytkowski: I see this is going to break. It re-implements checksum checks and uses the checksum dir. Need to fix it.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
rkorytkowski: I see what you mean. The problem is that the original algorithm doesn't delete checksums for config files that are no longer present. We don't do that with DB persisted checksums as well. I'm not sure if it should. The simple count comparison won't work.

api/src/main/java/org/openmrs/module/initializer/api/loaders/DispositionsLoader.java
rkorytkowski: No, I'll revert it.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
rkorytkowski: It's implemented in such a strange way... anyway I'm going to pass a map from getAllChecksums() to loadAddressConfiguration and have loadAddressConfiguration return a map with new or updated checksums to be persisted by the iniz module.

rkorytkowski: Fixed it! Changes in AddressHierarchy module committed in https://github.com/openmrs/openmrs-module-addresshierarchy/commit/b291b07286a8b10459fdbac0cd84d41194f3d771

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -76,23 +75,18 @@ public void started() {
...
-			boolean throwError = PROPS_STARTUP_LOAD_FAIL_ON_ERROR.equalsIgnoreCase(startupLoadingMode);
-			log.info("OpenMRS config loading process started...");
-			try {
-				getInitializerService().loadUnsafe(true, throwError);
-				log.info("OpenMRS config loading process completed.");
-			}
-			catch (Exception e) {
-				throw new ModuleException("An error occurred loading initializer configuration", e);
-			}
+			getInitializerService().runInitializer();
```
rkorytkowski: @mseaton moved the whole logic to runInitializer...

api/src/test/java/org/openmrs/module/initializer/api/AddressHierarchyLoaderIntegrationTest.java
```diff
@@ -92,7 +90,6 @@ public void load_shouldLoadAddressHierarchyMessages() throws IOException {
 		
 		// Replay
 		loader.load();
-		AddressConfigurationLoader.loadAddressConfiguration();
```
rkorytkowski, line 95: This line is actually handled by loader.load().

pom.xml
```diff
@@ -422,9 +422,6 @@
             <id>bahmni-nexus</id>
             <name>Bahmni Nexus Repository</name>
             <url>https://oss.sonatype.org/content/groups/public</url>
-            <snapshots>
-                <enabled>true</enabled>
-            </snapshots>
```
rkorytkowski, line 427: enabled by default

rkorytkowski:
I made all code review adjustments.

@dkayiwa and @ibacher I'll update refapp to use addreshierarchy 2.22.0-SNAPSHOT and new initializer version when merged.

rkorytkowski:
Tested with reference application. Migration from initializer 2.21.0 to 2.22.0-SNAPSHOT succeeded. Setup from scratch as well.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
```diff
@@ -2,15 +2,18 @@
...
 import org.openmrs.module.addresshierarchy.config.AddressConfigurationLoader;
 import org.openmrs.module.initializer.Domain;
 import org.openmrs.module.initializer.api.ConfigDirUtil;
+import org.openmrs.module.initializer.api.InitializerService;
+import org.openmrs.module.initializer.api.entities.InitializerChecksum;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@OpenmrsProfile(modules = { "addresshierarchy:2.17.* - 9.*" })
+@OpenmrsProfile(modules = { "addresshierarchy:2.22.* - 9.*" })
```
mseaton: This is not a good solution because everyone who is  running addresshierarchy version 2.17-2.21 will all of a sudden have their configurations no longer loaded by initializer if they upgrade, right?  We should try to retain compatibility and / or have 2 different versions of this Loader that target 2 different AH version ranges.

(Note:  I didn't see any PR to review for AH - did you just commit that directly?)

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
```diff
@@ -22,15 +25,15 @@ protected Domain getDomain() {
...
 	
 	@Override
 	public void loadUnsafe(List<String> wildcardExclusions, boolean doThrow) throws Exception {
-		
 		try {
-			AddressConfigurationLoader.loadAddressConfiguration(Paths.get(iniz.getConfigDirPath()),
-			    Paths.get(iniz.getChecksumsDirPath()));
+			Map<String, String> updatedChecksums = AddressConfigurationLoader
+			        .loadAddressConfiguration(Paths.get(iniz.getConfigDirPath()), iniz.getAllChecksums());
+			updatedChecksums.forEach((key, value) -> iniz.saveOrUpdateChecksum(new InitializerChecksum(key, value)));
```
mseaton: I'm not really sure this fix is needed here to be honest.  This _would_ be the way we would do it if initializer took responsibility over the addresshierarchy checksums and decision on whether to load or reload, but it doesn't.  We might (and probably do) want to move in a direction in which it does, but it doesn't currently.  Everything is managed in AH itself, it just happens to use the same `configuration` and `configuration_checksums` directory for consistency with the rest of the configuration.

What I would do is this:

- Keep the existing AH Loader supporting 2.17.* - 2.999.* and keep the exact same behavior as before.  If you have gotten rid of the `iniz.getChecksumsDirPath()`, then just make an equivalent within this loader that points to the same place as it did before for backwards compatibility
- Create a new AH Loader supporting AH version 3.* (3.0.0-SNAPSHOT is the master branch).  This should do what you are aiming to do, however I don't think you actually did things entirely correctly in AH yet, since AH has it's own logic for checksums and file loading, and you didn't do anything in AH to change that. I would think you actually need to make the same changes to checksum handing within AH that you are doing in Iniz _or_ you need to make AH fully depend on Iniz and have it delegate to the Iniz functionality.

(Note: I can see how all this might be a bit confusing since AH has files like ConfigUtil that are essentially copies from Iniz, but it is important to note that these are copies - AH does not currently depend on Iniz and has no knowledge of or ability to use it's functionality)

api/src/main/java/org/openmrs/module/initializer/api/ConfigDirUtil.java
```diff
@@ -89,8 +76,8 @@ public String getDomainDirPath() {
 		return domainDirPath;
 	}
 	
-	public String getDomainChecksumsDirPath() {
-		return domainChecksumsDirPath;
```
mseaton, line 93: Given there are changes to public API methods and constructors, we likely want to make this a new major version release of Iniz

api/src/main/java/org/openmrs/module/initializer/liquibase/DeleteDomainChecksumsChangeset.java
```diff
@@ -1,93 +0,0 @@
-package org.openmrs.module.initializer.liquibase;
```
mseaton, line 1: Why are we getting rid of this ChangeSet?  Because it isn't needed, or because we have changed how checksums are stored?  Because if it is the latter, that's incorrect.  We should just update this ChangeSet to delete the checksums from the new location.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -223,4 +248,256 @@ public InitializerConfig getInitializerConfig() {
...
+	public InitializerService getSelf() {
+		// If the service wasn't declared in xml, I would just use @Lazy on setSelf
+		return Context.getRegisteredComponent("initializer.InitializerService", InitializerService.class);
+	}
+	
+	@Override
+	public void runInitializer(String lockName) {
+		getSelf().migrateChecksumsFromFilesToDB();
+		
+		boolean throwError = PROPS_STARTUP_LOAD_FAIL_ON_ERROR
```
mseaton: I think we want this boolean to be passed into this method, and then for the activator to pass it in from the initializer config property, since that config is about what to do at startup vs. what to do when this method is called generally.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -223,4 +248,256 @@ public InitializerConfig getInitializerConfig() {
...
+		        .equalsIgnoreCase(getInitializerConfig().getStartupLoadingMode());
+		boolean lockAcquired = false;
+		
+		try {
```
mseaton: I don't know about all this.  If we do it this way, then any downstream code will need to be made aware of it.  Wouldn't it be better if the locking and checksum checking were to happen inside the existing `loadUnsafe` method(s)?  Then, if downstream modules are calling the `loadUnsafe` service methods directly, things will just continue to work?  Otherwise, things will not.

In our PIH distributions, we have found that we are better suited invoking initializer loading directly, as Iniz doesn't really handle everything the way that we would like.  See:  

https://github.com/PIH/openmrs-module-pihcore/blob/master/api/src/main/java/org/openmrs/module/pihcore/setup/InitializerSetup.java

Ideally, downstream code like this would either continue to work as expected, or would fail spectacularly either with startup failures or compilation errors.  Having things silently not do the right things because of these new changes would be non-ideal.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -76,23 +75,18 @@ public void started() {
 		InitializerMessageSource messageSource = getInitializerMessageSource();
 		Context.getMessageSourceService().setActiveMessageSource(messageSource);
 		
-		String startupLoadingMode = getInitializerService().getInitializerConfig().getStartupLoadingMode();
-		
+		String startupLoadingMode = getInitializerConfig().getStartupLoadingMode();
 		if (PROPS_STARTUP_LOAD_DISABLED.equalsIgnoreCase(startupLoadingMode)) {
 			log.info("OpenMRS config loading process disabled at initializer startup");
 		} else {
-			boolean throwError = PROPS_STARTUP_LOAD_FAIL_ON_ERROR.equalsIgnoreCase(startupLoadingMode);
```
mseaton, line 84: As I indicated above, we should continue to load this `throwError` property here, and pass it in

api/src/test/java/org/openmrs/module/initializer/api/loaders/BaseFileLoaderTest.java
```diff
@@ -67,158 +92,22 @@ public ConfigDirUtil getDirUtil() {
...
-			return null;
+		protected void load(File file) throws Exception {
 		}
 		
 		@Override
 		public String getDomainName() {
-			return "testdomain";
+			return "test";
 		}
-
```
mseaton, line 95: Are we sure all these tests are no longer needed?

api/src/test/java/org/openmrs/module/initializer/api/ConfigDirUtilTest.java
```diff
@@ -97,45 +89,6 @@ public void getFiles_shouldFilterOutBasedOnWildCardPattern() {
 		}
 	}
 	
-	@Test
```
mseaton, line 100: These tests still seem relevant?

api/src/test/java/org/openmrs/module/initializer/DomainBaseModuleContextSensitiveTest.java
```diff
@@ -71,7 +68,7 @@ public DomainBaseModuleContextSensitiveTest() {
 	
 	protected void initModules() {
 		{
-			Module mod = new Module("", "addresshierarchy", "", "", "", "2.17.0");
+			Module mod = new Module("", "addresshierarchy", "", "", "", "2.22.0");
```
mseaton: See earlier comments

pom.xml
```diff
@@ -92,7 +92,7 @@
         <serializationVersion>0.2.14</serializationVersion>
 
         <!-- For tests -->
-        <addresshierarchyVersion>2.17.0</addresshierarchyVersion>
+        <addresshierarchyVersion>2.22.0-SNAPSHOT</addresshierarchyVersion>
```
mseaton: Ditto

mseaton (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
mogoodrich: Sounds right, thanks @mseaton !

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
rkorytkowski: I see no way of supporting both versions easily. I would just create a loader for 2.17 - 2.21 that throws an error from a constructor saying you need to upgrade addresshierarchy. I linked the direct commit in AH from this PR.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
rkorytkowski: I see where you are going with supporting old versions of AddressHierarchy. I would then need to exclude the addresshierarchy domain from migrating to checksums stored in DB if addresshierarchy is not 2.22+ and do migrate as soon as you upgrade... It gets messy in my opinion. I would just ask to upgrade to newer adresshierarchy.

AH logic is here https://github.com/openmrs/openmrs-module-addresshierarchy/commit/1bbf1da5165cb863fa5467e506820c9548eb62e6
It sill lets AH handle checksums in any way it wants including how it calculates checksums, but delegates the loading of checksums and persistence to the initializer module. The way things are architected is that, initializer is made aware of addresshierarchy to be started after addresshierarchy so I cannot make adress hierarchy depend on initializer...

api/src/main/java/org/openmrs/module/initializer/api/ConfigDirUtil.java
rkorytkowski: Sounds good to me!

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
rkorytkowski: If you mean moving things to InitializerServiceImpl.loadUnsafe(boolean applyFilters, boolean doThrow), I'm fine with that. I wouldn't move locking to BaseLoader.loadUnsafe as then each node would compete to run BaseLoader.loadUnsafe and it would be a nightmare to monitor and debug the initialization process.

api/src/main/java/org/openmrs/module/initializer/liquibase/DeleteDomainChecksumsChangeset.java
rkorytkowski: It's a very old changeset from 2021... I'm fine keeping it around or deleting. Definitely no use in DB checksums. I would need to keep the old ConfigDirUtil.deleteChecksums method if it is to stay.

api/src/test/java/org/openmrs/module/initializer/api/loaders/BaseFileLoaderTest.java
rkorytkowski: I'll update them to DB checksums.

api/src/test/java/org/openmrs/module/initializer/api/ConfigDirUtilTest.java
rkorytkowski: We have subdirectories covered in migration tests, but missing in DB checksums. I'll add this case to BaseFileLoaderTest.java‎.

rkorytkowski:
@mseaton I think I responded to all your comments now. Looking forward to your reply on deleting the old changeset, addresshierarchy changes and locking for loadUnsafe.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
mseaton: The problem is, there is no way for Iniz to indicate that it no longer supports versions 2.17 - 2.21.  It only has an aware-of dependency, not a require module dependency.  So, if we can't support it, we should at least throw an Exception if someone tries to use the loader with versions that were supported and now are no longer supported, so that it is obvious that behavior no longer works and they need to take action.

It would also make sense to ensure the version associated with this PR is a new major version of Iniz, to reflect this loss of compatibility

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
mseaton: Given the way that Iniz has evolved in the years since AH first created this mechanism, we are long overdue to remove the loading functionality from AH and move it entirely into Iniz, which is what I think we should do here.

If your overall goal is for the refapp platform to support horizontal scaling, addresshierarchy will need much the same changes that you are making to Iniz anyway.

What I think we should do is:

* Create a new major version release of AH.  In this version:
** Remove all checksum-related logic from AH, and also remove any knowledge of the "configuration" and "configuration_checksums" directories altogether, and just keep the functionality that loads/reloads data from configuration files if invoked.  
** Remove the functionality from the AH activator to load address files at startup
* In Iniz:
** Add a Processor with an OpenMRS profile that targets this new version of AH
** Move the logic from AH to this new processor as appropriate.

api/src/main/java/org/openmrs/module/initializer/api/ConfigDirUtil.java
mseaton: Great - I don't see that in this PR yet?

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
mseaton: Not sure I follow.  Different nodes will process different domains during the same loading cycle?

mseaton (changes requested):

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
rkorytkowski: It's why I suggested having a loader for 2.17-2.21 throwing an error and asking to upgrade. I'll add that.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
rkorytkowski: Okay, so you would like me to refactor and move the entire initialisation logic from AH to the iniz module. I think it's how it should have been implemented from the beginning. It was a wired construct with AH handling its own checksums and be called by iniz. I thought it was for a reason. I didn't want to make more changes than needed in AH to support persisting checksums in DB.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
rkorytkowski: If I am to do what you are asking for i.e. "the locking and checksum checking were to happen inside the existing loadUnsafe method(s)" then each loadUnsafe method will be trying to get a lock to run and they will be queued up to run by different nodes competing to get that lock.

The lock must be hold for the whole initialisation to happen on a single node, thus I hold it for the whole initialisation in runInitializer. I can move the logic to InitializerServiceImpl.loadUnsafe(boolean applyFilters, boolean doThrow) which calls individual loadUnsafe methods in loader, if it helps, but not sure it changes much. You would still need to refactor your InitializerSetup.java to call InitializerServiceImpl.loadUnsafe

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
mseaton: I think that would be best at this point.  My recollection is that we built this feature in AH around 10 years ago, and realized that it would be a really good general purpose solution and so @mks-d and @rbuisson ran with it and the initializer module was born.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
mseaton: At the end of the day, as long as sufficient methods are exposed publicly from Iniz that allows me to manage loading the domains that I want, when I want, and I can wrap that in the necessary calls to lock, etc, then that should work.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader_2_17.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.initializer.Domain;
+
+@OpenmrsProfile(modules = { "addresshierarchy:2.17.* - 2.21.*" })
+public class AddressHierarchyLoader_2_17 extends BaseLoader {
+	
+	public AddressHierarchyLoader_2_17() {
+		throw new IllegalStateException("You must upgrade addresshierarchy to 2.22+.");
```
rkorytkowski: No easy way to write a test for it, but it's enough to change version in DomainBaseModuleContextSensitiveTest to see the app failing to load.

api/src/main/java/org/openmrs/module/initializer/api/ConfigDirUtil.java
rkorytkowski: Done.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
rkorytkowski: Done.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
rkorytkowski: Done.

api/src/main/java/org/openmrs/module/initializer/InitializerActivator.java
```diff
@@ -76,23 +76,27 @@ public void started() {
...
+		String startupLoadingMode = getInitializerConfig().getStartupLoadingMode();
 		if (PROPS_STARTUP_LOAD_DISABLED.equalsIgnoreCase(startupLoadingMode)) {
 			log.info("OpenMRS config loading process disabled at initializer startup");
 		} else {
 			boolean throwError = PROPS_STARTUP_LOAD_FAIL_ON_ERROR.equalsIgnoreCase(startupLoadingMode);
-			log.info("OpenMRS config loading process started...");
+			
 			try {
 				getInitializerService().loadUnsafe(true, throwError);
-				log.info("OpenMRS config loading process completed.");
```
rkorytkowski, line 88: Moved info messages to loadUnsafe

api/src/test/java/org/openmrs/module/initializer/api/loaders/BaseFileLoaderTest.java
rkorytkowski: Done.

api/src/test/java/org/openmrs/module/initializer/api/ConfigDirUtilTest.java
rkorytkowski: Done in getFiles_shouldReturnNestedFiles and loadUnsafe_shouldHandleNestedFiles

rkorytkowski:
@mseaton addressed all your review comments.

I didn't make any changes to AH. No AH upgrade is needed as well. In iniz 3.0.0 all the AH loading is done in the iniz module.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
```diff
@@ -40,4 +66,247 @@ public void loadUnsafe(List<String> wildcardExclusions, boolean doThrow) throws
 			}
 		}
 	}
+	
+	public void loadAddressConfiguration(Path configPath, Map<String, String> previousChecksums) {
```
rkorytkowski, line 59: All below is copied from AddressConfigurationLoader with just a few lines changed to use DB checksums instead of files.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
```diff
@@ -1,19 +1,47 @@
...
+import org.openmrs.util.OpenmrsConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * The logic is a copy from
+ * {@link org.openmrs.module.addresshierarchy.config.AddressConfigurationLoader} except for
+ * checksums loading and persisting.
+ */
 @OpenmrsProfile(modules = { "addresshierarchy:2.17.* - 9.*" })
```
rkorytkowski, line 28: No AH upgrade needed.

rkorytkowski:
@rbuisson It would be great to have your team review it as well.

api/src/main/java/org/openmrs/module/initializer/api/loaders/AddressHierarchyLoader.java
```diff
@@ -40,4 +66,247 @@ public void loadUnsafe(List<String> wildcardExclusions, boolean doThrow) throws
...
+			installAddressHierarchyEntries(configUtil, addressConfiguration.getAddressHierarchyFile(),
+			    forceReloadEntries || addressConfiguration.mustWipe());
+			iniz.saveOrUpdateChecksum(checksum);
+			
+			log.info("Entries loaded, re-initializing address cache");
+			getService().initializeFullAddressCache();
+		}
+		getService().initI18nCache();
+	}
+
```
mseaton: I would think that any/all of the methods below that are unrelated to checksums and which are already present in the AH module should just be used directly from the AH module, and not re-copied into here.  Assuming these methods are adequately maintained in the AH module, and may be useful to use independently on the Initialilzer module, I don't see why we would duplicate/copy them into here?

rkorytkowski: I'm very confused with this AH module altogether. Probably not the best person to do refactoring there. I understand you want to get rid of technical debt there.

I don't know why AH calls loadAddressConfiguration in its activator and then iniz calls loadAddressConfiguration again... Is the activator call disabled in some way when you have iniz installed?

ibacher: Yeah, @mseaton I'm unsure how to interpret this. The AH re-running `loadAddressConfiguration()` is safe when Iniz is generating the same checksum files because the checksums should match. But it seems like _either_ we're moving the initialization stuff into Iniz or we shouldn't have Iniz worry about AH at all.

Keeping the initialization logic in AH seems to require that AH depend on Iniz, but that's the exact opposite of the paradigm we've adopted for Iniz. I think the sane thing to do here is move all of the initialization parts into Iniz and remove them from AH. But I'm curious how you're envisioning things working?

mseaton: @ibacher / @rkorytkowski - all I'm saying is that if there are existing methods in AH that don't deal with checksums, but just take responsibility for the actual loading of a config file, then we should just use the existing.  Why are we copying methods like:

`public static void installAddressHierarchyLevels(List<AddressComponent> addressComponents)`
or
`AddressConfiguration readFromString(String configuration)`
or
`String writeToString(AddressConfiguration configuration)`
or
`XStream getSerializer()`
or
`void setGlobalProperty(String propertyName, String propertyValue)`

etc, etc. ?

Ideally AH would retain the capabilities to delete / load Address Level and Address Hierarchy information from appropriate files, and Iniz is just invoking that functionality based on configuration files located in the configuration directory and if they are deemed to have changed based on previous checksums.

mseaton: > I don't know why AH calls loadAddressConfiguration in its activator and then iniz calls loadAddressConfiguration again... Is the activator call disabled in some way when you have iniz installed?

Legacy.  I think we just never noticed this because of the way the checksums work - i.e. if it is loaded by Iniz, then when the AH activator tries to load it again it will see the checksums have not changed and so won't load it again.

rkorytkowski: Okay, so in the current approach (in this PR) no changes to AH are needed (except for fixing the AH startup procedure to not load address hierarchy if iniz is installed). I can call all public methods from the AH AddressConfigurationLoader instead of copying them to the iniz module except for loadAddressConfiguration. I keep a copy of loadAddressConfiguration in the Iniz module and modify it to use DB checksums and call public methods from AH AddressConfigurationLoader. Would that work?

Or do you want me to refactor loadAddressConfiguration in the AH module to not do any checksum checks and simply import given xml and csv. I would probably need to divide the loading into 2 separate methods (one for xml and one for csv) to be called from the Iniz module. The iniz module needs to know, which xml and csv files to check and pass to AH to load. In this approach the AH needs to be upgraded to work with the 3.0.x iniz module.

mseaton: > I can call all public methods from the AH AddressConfigurationLoader instead of copying them to the iniz module except for loadAddressConfiguration. I keep a copy of loadAddressConfiguration in the Iniz module and modify it to use DB checksums and call public methods from AH AddressConfigurationLoader. Would that work?

Yes, this is all I am saying.  If there is an existing method in AH that can be used as-is, just use it, don't copy it over from there unnecessarily.

rkorytkowski: Done!

rkorytkowski:
@mseaton @kipchubett @ibacher please let me know if you still see any issues with this change

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -223,4 +284,224 @@ public InitializerConfig getInitializerConfig() {
...
+	public InitializerService getSelf() {
+		// If the service wasn't declared in xml, I would just use @Lazy on setSelf
+		return Context.getRegisteredComponent("initializer.InitializerService", InitializerService.class);
+	}
+	
+	@Transactional
+	@Override
+	public void migrateChecksumsFromFilesToDB() {
+		String globalProperty = adminService.getGlobalProperty("initializer.checksumsMigrated");
+		if (globalProperty == null) {
```
kipchubett: If the global property exists but is set to "", "false", or any value other than `null` (there's a chance that someone might mess up the settings by mistake), the check will fails thus migration won't run. It may be better to explicitly check for `true`.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
```diff
@@ -223,4 +284,224 @@ public InitializerConfig getInitializerConfig() {
...
+			// Migrate file checksums, to be removed in later versions
+			Path checksumsBase = Paths.get(getChecksumsDirPath());
+			Path base = Paths.get(getConfigDirPath());
+			try (Stream<Path> stream = Files.walk(base)) {
+				stream.filter(Files::isRegularFile).forEach(path -> {
+					String relConfigFilePath = base.relativize(path).toString();
+					//Replace all '/' with '_' except for the first one separating domain
+					relConfigFilePath = StringUtils.substringBefore(relConfigFilePath, File.separator) + File.separator
+					        + StringUtils.substringAfter(relConfigFilePath, File.separator).replace("/", "_");
```
kipchubett: On Windows, paths use `\`, possibly this will not work for Windows users.

kipchubett (commented):
Generally, looks good. I've left some comments.

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
rkorytkowski: Addressed, thanks!

api/src/main/java/org/openmrs/module/initializer/api/InitializerServiceImpl.java
rkorytkowski: Addressed, thanks!

dkayiwa:
@rkorytkowski did you take a look at the build failure?

rkorytkowski:
@dkayiwa they didn't run for me and I had no permissions to trigger them... Just checked and the failure doesn't seem to be related to this PR. Do you have permissions to run tests on the main branch to verify if they still pass? They haven't run for a month... I'm not able to reproduce the failure locally.

dkayiwa:
@kipchubett do you have permissions to run those tests?

rkorytkowski:
@dkayiwa I'd also try re-triggering tests on this PR. I don't have permissions to do that either.

kipchubett:
I've re-triggered on main branch here: https://github.com/mekomsolutions/openmrs-module-initializer/actions/runs/22675174310/job/69163801641
It does fail on main as well, same error. We will need to be fix it.

dkayiwa:
@kipchubett was it caused by the changes in this pull request? Or is it a failure that already existed? In other words, do the changes in this pull requests compile successfully when you try them out locally?

dkayiwa:
@kipchubett are we then ready to merge this pull request?

kipchubett:
> @kipchubett are we then ready to merge this pull request?

I'll do some further tests to figure out the issue. It doesn't seem to be related with this PR

dkayiwa:
@kipchubett is there anything left before you can merge this?

kipchubett:
@dkayiwa @rkorytkowski This seems fine to me. We can merge it in. @rkorytkowski could you update this branch with the latest `main`

rkorytkowski:
@kipchubett done! I'm sorry for the delay. I was away.

PR #324, opened 2026-05-22
O3-5702: Add support for Cashier Item Prices domain in the Initializer module

wikumChamith:
This adds new cashieritemprices CSV domain: loader, CSV parser, line processor.


wikumChamith:
@ibacher @NethmiRodrigo could you please review this PR?

api-2.7/src/main/java/org/openmrs/module/initializer/api/billing/CashierItemPriceLineProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+			throw new IllegalArgumentException("Exactly one of '" + HEADER_STOCK_ITEM + "' or '" + HEADER_BILLABLE_SERVICE
+			        + "' must be set on each row");
+		}
+		
+		if (hasStockItem) {
+			StockItem stockItem = stockManagementService.getStockItemByUuid(stockItemUuid);
+			if (stockItem == null) {
+				throw new IllegalArgumentException("No StockItem found with UUID '" + stockItemUuid + "'");
+			}
+			instance.setItem(stockItem);
```
denniskigen, line 72: When updating an existing `CashierItemPrice` from one association type to the other, this keeps the old association around. For example, if the existing row has a `billableService` and the CSV now supplies a stock item, line 72 sets `item` but `billableService` remains set; the reverse happens in the billable service branch. I validated this with a temporary unit test against `fill()`, and both transitions left the old association non-null.

Could we clear the opposite field in each branch, e.g. `setBillableService(null)` when setting `item` and `setItem(null)` when setting `billableService`, and add a regression test for switching association type?

api-2.7/src/main/java/org/openmrs/module/initializer/api/billing/CashierItemPriceLineProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+import org.openmrs.module.billing.api.model.BillableService;
+import org.openmrs.module.billing.api.model.CashierItemPrice;
+import org.openmrs.module.billing.api.model.PaymentMode;
+import org.openmrs.module.initializer.api.BaseLineProcessor;
+import org.openmrs.module.initializer.api.CsvLine;
+import org.openmrs.module.stockmanagement.api.StockManagementService;
+import org.openmrs.module.stockmanagement.api.model.StockItem;
+import org.springframework.beans.factory.annotation.Autowired;
+
+@OpenmrsProfile(modules = { "billing:2.0.0 - 9.*" })
```
NethmiRodrigo, line 18: Only billing is declared here, but shouldn’t stock management be declared too?

api-2.7/src/main/java/org/openmrs/module/initializer/api/billing/CashierItemPriceLineProcessor.java
wikumChamith: @denniskigen I fixed this. This is actually an architectural issue in the billing module. CashierItemPrice is a sum type (a price for either a BillableService or a StockItem), and modelling it as one entity with two nullable FK columns is awkward in the first place. A better option would be to use a base abstract CashierItemPrice (or rename it ItemPrice) with two concrete subclasses:

```
- CashierItemPrice (abstract)
- ServiceItemPrice extends CashierItemPrice
- StockItemPrice extends CashierItemPrice
```

This is also more scalable as we can easily add more types if needed. This is something we should look into in the future, though it would be a little tricky to do without breaking existing instances :)

api-2.7/src/main/java/org/openmrs/module/initializer/api/billing/CashierItemPriceLineProcessor.java
wikumChamith: Billing has `stockmanagement` 1.4.0+ as a hard `<require_module>`, so the billing literally cannot start without stockmanagement present. So wherever `billing:2.0.0 - 9.*` is satisfied, `stockmanagement` is guaranteed to be there too. Adding it here would be redundant.

wikumChamith:
@ibacher is this good to get merged?

readme/cashieritemprices.md
```diff
@@ -0,0 +1,44 @@
...
+The price amount, parsed as a `BigDecimal` (e.g. `150.00`).
+
+###### Header `Payment Mode` *(required)*
+The UUID of an existing `PaymentMode`.
+
+###### Header `Stock Item` *(required if `Billable Service` is empty)*
+The UUID of an existing stockmanagement `StockItem`. The Initializer module does not load stock items; the referenced stock item must already exist in the database (for example, created through the stockmanagement module's UI).
+
+###### Header `Billable Service` *(required if `Stock Item` is empty)*
+The UUID of an existing `BillableService`. See [Billable Services](billableservices.md).
```
denniskigen, line 37: Tiny docs nit: the domain supports retiring rows via a `Void/Retire` column (the test CSV uses it), but it isn't listed among the headers here. Should we add a short entry for it so the docs cover the full header set?

denniskigen (approved):
Thanks, @wikumChamith. LGTM.

api-2.7/src/test/resources/testAppDataDir/configuration/cashieritemprices/cashierItemPrices.csv
```diff
@@ -0,0 +1,4 @@
+Uuid, Void/Retire, Name, Price, Payment Mode, Stock Item, Billable Service
+c1c1c1c1-0000-0000-0000-000000000001,, ANC Service Price, 150.00, 526bf278-ba81-4436-b867-c2f6641d060a,, a0f7d8a1-4fa2-418c-aa8a-cccccccc0001
+c1c1c1c1-0000-0000-0000-000000000002,, Ortho Service Price (Modified), 250.50, 2b1b9aae-5d35-43dd-9214-3fd370fd7737,, a0f7d8a1-4fa2-418c-aa8a-cccccccc0001
+c1c1c1c1-0000-0000-0000-000000000003, true, Nutrition Service Price, 75.00, e168c141-f5fd-4eec-bd3e-633bed1c9606,, a0f7d8a1-4fa2-418c-aa8a-cccccccc0001
```
denniskigen: This retire row references payment mode `e168c141-f5fd-4eec-bd3e-633bed1c9606`, which isn't defined in the test setup or used elsewhere. It passes only because `shouldFill()` skips `fill()` for existing rows marked for retirement, so the UUID is never resolved. Should we reuse one of the real payment mode UUIDs or is that not strictly necessary?

denniskigen:
Thanks @wikumChamith. Nothing further from me, this looks good from my side.

ibacher (approved):

PR #331, opened 2026-07-17
O3-5820: Fix OrderType parent silently dropped on import for concrete order types

wikumChamith:
Utils.getParentOrderType now resolves the parent for any order type, not only those whose Java class name is the literal org.openmrs.Order


ibacher (approved):
But can you add a line to the README changelog for this?

wikumChamith:
> But can you add a line to the README changelog for this?

@ibacher done.

